Issue #4, opened 2019-02-26
Improve Windows build process


uranusjr:
A couple of things I have in mind, will send a PR if I am successful in implementing them:

* Use [Strawberry Perl](http://strawberryperl.com/releases.html) to build OpenSSL, instead of depending on the `PERL` environment variable. Strawberry Perl has a nice portable distribution.
* Automatically set environment variables *a la* distutils, so I don’t forget to run `vcvarsall.bat` before building.

indygreg:
These improvements are very much appreciated!

The Windows build is the least tested of the 3 and definitely the hackiest due to Visual Studio project file rewriting. (Linux is most tested.)

I spent way too many hours trying to get the Windows build working. When it finally reached a point that I was able prove the viability of PyOxidizer on Windows, I took a break. Depending on how far down the rabbit hole you go, you may find that various build settings aren't yet correct for standalone binaries. e.g. I believe we still need to tweak which C runtime to use and there may be some lingering DLL dependencies that can be eliminated.

But improvements to the build to make it more turnkey are very much welcome!

indygreg:
This issue doesn't appear actionable, so closing.

PR #5, opened 2019-02-26
Use Strawberry Perl to build OpenSSL

uranusjr:
First part of #4, plus various bug fixes necessary for me to build the end product 😄


indygreg:
I manually rebased and pushed these commits.

Thank you so much for automating the Perl dependency!

PR #20, opened 2019-11-20
windows: Rename venv to venv.windows

jayvdb:
Related to https://github.com/indygreg/python-build-standalone/issues/19


indygreg:
The reason `venv` isn't included is because `--include-venv` isn't being passed to the tool used to install files. I have a proper fix for this already committed in my local repo.

PR #61, opened 2020-10-28
Took care of the bad shebangs quirk

zsimic:
Make all python scripts in `install/bin` portable as well


zsimic:
I've tested this BTW -> all python wrappers in `install/bin` work as expected, no more quirk :)

zsimic:
Would it be possible to merge this PR?

indygreg:
92fe187a43f77a6a5e03581f4944225b94b1b328 implements a rewriting of the shebangs. The implementation is a bit different from this PR. But it was inspired by it and credited in the commit message (and by this comment).

This PR is no longer needed.

PR #93, opened 2021-10-01
WIP: Windows arm64 builds - help needed

adiantek:
Hello,

I'm trying to add Windows ARM64 compatibility in your python-build-standalone. I started working with it. OpenSSL is compiled, libffi checks done, but 
https://github.com/adiantek/python-build-standalone/runs/3770671733?check_suite_focus=true#step:7:7470 why `-march64` is ignored?

And I don't know why build system is 32-bit... Build system should be `x86_64-w64-cygwin`, and host/target aarch64:
```
checking build system type... i686-pc-cygwin
checking host system type... aarch64-w64-cygwin
checking target system type... aarch64-w64-cygwin
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -43,8 +43,9 @@ jobs:
           - 'cpython-3.9'
           - 'cpython-3.10'
         vcvars:
-          - 'vcvars32.bat'
+          - 'vcvarsamd64_x86.bat'
           - 'vcvars64.bat'
+          - 'vcvarsamd64_arm64.bat'
         profile:
           - 'static-noopt'
           - 'shared-pgo'
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -43,7 +43,7 @@ jobs:
           - 'cpython-3.9'
           - 'cpython-3.10'
         vcvars:
-          - 'vcvarsamd64_x86.bat'
+          - 'vcvars32.bat'
           - 'vcvars64.bat'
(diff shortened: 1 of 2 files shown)


indygreg:
The problem here is that the `msvc.sh` script in the version of libffi we are using doesn't recognize the `-arm64` argument. It looks like the newer source of libffi in https://github.com/python/cpython-source-deps/tree/libffi does support this.

We should upgrade to the latest libffi and in a separate commit/PR add support for arm/arm64 builds. I'll start looking into upgrading libffi now.

indygreg:
I pushed a commit to the `main` branch to update libffi to version 3.4.2 on Windows. After this change, I'm able to build an arm64 libffi. So you should be unblocked from finishing this PR. (It currently fails in CPython's build when handling OpenSSL when I build locally.)

Also, if we're going to implement ARM64, what do you think about doing ARM as well? I feel like it should be trivial to do both once you get 1 working? If you don't want to deal with the scope bloat, I can work on ARM once you get ARM64 working.

zooba:
> Also, if we're going to implement ARM64, what do you think about doing ARM as well? I feel like it should be trivial to do both once you get 1 working? If you don't want to deal with the scope bloat, I can work on ARM once you get ARM64 working.

Only do this if you're keen - it seems like ARM is not going to be an important user or server platform anytime soon, and the IoT applications aren't as compelling as they once were. ARM64 is the main priority for everyone.

zanieb:
I'm continuing this work over in https://github.com/astral-sh/python-build-standalone/pull/387

PR #101, opened 2021-11-10
windows: restore windows 7 support

silverjam:
This restores windows 7 support just for Python 3.8 by making the `pathcch` library conditional.

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -1869,15 +1869,15 @@ def collect_python_build_artifacts(
         {"name": "shlwapi", "system": True},
         {"name": "version", "system": True},
         {"name": "ws2_32", "system": True},
-        # pathcch is Windows 8+ only. Python 3.9 dropped support for Windows 7.
-        # So this dependency is technically incorrect on Python 3.8.
-        {"name": "pathcch", "system": True},
         # In addition to the ones explicitly in the project, there are some
         # implicit link libraries not present. We list those as well.
         {"name": "Ole32", "system": True},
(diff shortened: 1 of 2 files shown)


cpython-windows/build.py
```diff
@@ -1869,16 +1869,16 @@ def find_additional_dependencies(project: pathlib.Path):
...
-        # So this dependency is technically incorrect on Python 3.8.
-        {"name": "pathcch", "system": True},
         # In addition to the ones explicitly in the project, there are some
         # implicit link libraries not present. We list those as well.
         {"name": "Ole32", "system": True},
         {"name": "OleAut32", "system": True},
         {"name": "User32", "system": True},
     ]
 
+    if python_majmin in ["39", "310"]:
```
silverjam, line 1879: Not sure how to make this more future proof, it would be nice to say "if greater than Python 3.9"

indygreg: We're not supporting any versions older than 3.8. So changing this to `if python_majmin != "38":` should do the job. I'll change this in flight.

indygreg (approved):
Restoring compatibility with Windows 7 for Python 3.8 seems reasonable to me. I'll apply some minor tweaks in flight and merge this via a rebase outside of GitHub.

Thank you for contributing this change.

silverjam:
> Restoring compatibility with Windows 7 for Python 3.8 seems reasonable to me. I'll apply some minor tweaks in flight and merge this via a rebase outside of GitHub.
> 
> 
> 
> Thank you for contributing this change.

Great, thanks!

PR #102, opened 2021-12-17
fix: don't install openssl docs

f0rmiga:
The installation of the openssl docs is incredibly long. The change set presented here avoids the installation of the docs.


indygreg (approved):
Yes, the OpenSSL docs installation is absurdly long for what it is doing. This has long bothered me but I never took the time to look into disabling the docs.

The `make install` target literally targets the 2 targets you added here plus a docs target. So this change looks safe to me.

Thank you for contributing this!

f0rmiga:
My pleasure. Thanks for your effort in keeping this repo well-maintained.

PR #105, opened 2022-01-11
Update Python 3.10 to 3.10.1

caseyduquettesc:
Not sure what tests you typically run. It at least built fine on mac amd64/aarch4 and linux amd64/aarch64.

```
./build-macos.py --optimizations pgo --target-triple x86_64-apple-darwin --python cpython-3.10
./build-macos.py --target-triple aarch64-apple-darwin --python cpython-3.10

./build-linux.py --optimizations pgo --target x86_64-unknown-linux-gnu --python cpython-3.10
./build-linux.py --optimizations pgo --target aarch64-unknown-linux-gnu --python cpython-3.10
```

I have a bazel toolchains project with a python test suite I'll run against these builds and report back if they look ok.

I ran our test suite, which basically consists of
* `pip install pycryptodome==3.4.11` to confirm native extension compilation works
* checking `platform.python_version()` equals 3.10.1
* and some basic functionality tests of `multiprocessing`, `ctypes`, `uuid`, `lzma`, `socket`, `ssl`, and `asyncio`

All 4 interpreters passed. Can't validate the windows builds unfortunately and there's a TODO related to python 3.10.1 + windows I presume will need to be taken care of too.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -67,7 +67,7 @@ if [ "${BUILD_TRIPLE}" != "${TARGET_TRIPLE}" ]; then
 
   # Same patch as below. See comment there.
   if [ "${CC}" = "clang" ]; then
-    if [ "${PYTHON_MAJMIN_VERSION}" = "3.9" ]; then
+    if [ "${PYTHON_MAJMIN_VERSION}" = "3.9" ] || [ "${PYTHON_MAJMIN_VERSION}" = "3.10" ]; then
       patch -p1 <<"EOF"
 diff --git a/configure b/configure
 index 33ecb16f71..7f21822d97 100755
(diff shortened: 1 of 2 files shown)


indygreg:
Thank you for this PR and for reporting your validation!

In the time between you submitting this and me reviewing it, 3.10.2 was released. So we should upgrade to that.

Also, it looks like the `MULTIARCH` configure code changed in 3.10.2 and 3.9.10, so the upgrade isn't as trivial as I'd like. I hope you don't mind, but I'm going to close this in favor of upgrading straight to 3.9.10 and 3.10.2 and addressing the configure changes. I'll hopefully merge that change within a few hours.

caseyduquettesc:
@indygreg I agree with your decision. Thanks for the updates and thanks for the project. It's quite wonderful

indygreg:
I think I have the 3.10.2 upgrade patch working. Just waiting on some final CI runs to confirm. Hopefully I push it to the `main` branch within a few hours.

Closing this PR per discussion.

PR #112, opened 2022-02-15
unix: wrap sccache call with HAVE_SCCACHE

DanielShaulov:
Despite checking for the existence of sccache, there is a naked call to it not wrapped with the check for `HAVE_SCCACHE`, added the check


indygreg (approved):
Thanks for catching this!

PR #120, opened 2022-03-29
Fix zlib download URL

ddeville:
zlib just released 1.2.12 on 03/22 and they only have the latest version available under zlib.net (so right now it's
https://www.zlib.net/zlib-1.2.12.tar.gz and https://www.zlib.net/zlib-1.2.11.tar.gz 404s and the Python build is broken.

However, they keep all versions (including the latest) under zlib.net/fossils so use that path instead.


indygreg (approved):
Thanks!

PR #157, opened 2023-01-18
unix: fix CLFAGS typo in some scripts.

chenxu2048:
We tried to build python with aarch64 and musl, but got failed in some packages.

These packages use CLFAGS while building and some substantial flags for musl-clang like `--target=aarch64-linux-musl` can not be read by configure scripts.

NOTE: fix may cause failed in some variants building.

### DIFF
--- a/cpython-unix/build-gdbm.sh
+++ b/cpython-unix/build-gdbm.sh
@@ -15,7 +15,7 @@ pushd gdbm-${GDBM_VERSION}
 
 # CPython setup.py looks for libgdbm_compat and gdbm-ndbm.h,
 # which require --enable-libgdbm-compat.
-CLFAGS="${EXTRA_TARGET_CFLAGS} -fPIC" CPPFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC" ./configure \
+CFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC" CPPFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC" ./configure \
     --build=${BUILD_TRIPLE} \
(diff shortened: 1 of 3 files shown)


indygreg (approved):
Oh yikes. Thank you for this PR!

I'll merge this once CI comes back green. I don't anticipate any failures. But you never know.

Thankfully our exposure to this bug should be small.

GDBM isn't enabled any more. patchelf is always targeting the current/host machine, so no additional compiler flags should be needed.

That leaves libedit.

It looks like we weren't including `EXTRA_TARGET_CFLAGS` in `CFLAGS` but were in `CPPFLAGS`. This likely broke cross-compilation on Apple and some Linux builds. The Linux targets not using our custom Clang toolchain were likely fine, as we explicitly point at a GCC toolchain that cross-compiles by default. But if we were using the custom Clang toolchain, that targets the host by default so unwanted host state (including the default macOS SDK target) may have slipped into the built binaries.

This is probably serious enough to warrant a new release.

Although if this is a problem, I'm shocked nobody has complained. I would be expecting this to break linking or segfault at runtime if we're mixing architectures. Strange.

indygreg:
OK. We use the system libedit on macOS. So exposure here is to Linux.

The Rust validation code (which runs in CI) validates the ELF and Mach-O headers to ensure the advertised CPU type matches what is expected for the target triple. I spot checked that this was correctly inspecting all `.o` files, even though inside `.a` archives and that it was looking at libedit's `.o` files. It was.

So I think that means none of the distributions we have been`` shipping are impacted by this. i.e. it is only a bug for local cross-compile scenarios.

PR #181, opened 2023-07-21
Update the patchelf version: 0.12 -> 0.13.1 to fix corruption on s390x

jonathan-albrecht-ibm:
downloads: patchelf 0.12 -> 0.13.1

patchelf 0.13.1 fixes a bug that caused the python3* executable to be corrupted on s390x. Attempting to run python3 on an s390x machine would produce an error like:

install/bin/python3: Permission denied

patchelf 0.13.1 is the most recent version that can be compiled by the debian stretch gcc since newer versions of patchelf require c++-17.

I have access to x86_64, s390x and ppc64le VMs so I did a quick build and smoke test of python 3.11 on those platforms and they looked good. I believe only the install/bin/python3* executables are processed by patchelf so I think my local testing should provide good coverage but let me know if there is anything specific I should try.

### DIFF
--- a/cpython-unix/build-patchelf.sh
+++ b/cpython-unix/build-patchelf.sh
@@ -11,7 +11,7 @@ export PATH=/tools/${TOOLCHAIN}/bin:/tools/host/bin:$PATH
 
 tar -xf patchelf-${PATCHELF_VERSION}.tar.bz2
 
-pushd patchelf-0.12.20200827.8d3a16e
+pushd patchelf-0.13.1.20211127.72b6d44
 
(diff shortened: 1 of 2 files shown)


jonathan-albrecht-ibm:
@indygreg I just wanted to note that this is the PR mentioned in https://github.com/indygreg/python-build-standalone/pull/180#issuecomment-1641404600

indygreg (approved):
This is not the s390x compatibility patch I was expecting! I'm honestly surprised patchelf was failing on just this platform!

Thank you for the fix. Will merge it once CI comes back green. (Which may take a few retries since we've been having issues with snapshot.debian.org reliability recently.)

PR #215, opened 2024-02-19
Fix "Cannot read termcap database;"

orent:
./configure with defaults search path compatible with both FHS and the soon-to-be-deprecated Debian path

https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#idm236087499712

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1028202

### DIFF
--- a/cpython-unix/build-ncurses.sh
+++ b/cpython-unix/build-ncurses.sh
@@ -20,7 +20,7 @@ if [[ -n "${CROSS_COMPILING}" && "${PYBUILD_PLATFORM}" != "macos" ]]; then
   echo "building host ncurses to provide modern tic for cross-compile"
 
   pushd ncurses-${NCURSES_VERSION}
-  CC="${HOST_CC}" ./configure --prefix=${TOOLS_PATH}/host --without-cxx --without-tests --without-manpages --enable-widec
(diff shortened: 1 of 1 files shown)


indygreg (commented):
I support the intent of the change.

But the diff as authored is changing the temporary host build of terminfo, not the final build we actually use.

If you go look later in the file, you'll see that we set the following:

```
--with-terminfo-dirs=/usr/share/terminfo
--with-default-terminfo-dir=/usr/share/terminfo
```

I think the problem here is that older Debian versions are putting stuff in `/lib/terminfo`.

Looking at the code in PyOxidizer, I think `/etc/terminfo:/lib/terminfo:/usr/share/terminfo` might be a reasonably safe default, as that should be compatible with both Debian and RedHat flavored distros.

I'll cherry pick and amend this commit to make the changes.

orent:
RHEL, MacOS, and any other unix descentant other than Debian and its derivatives install terminfo data in /usr/share/terminfo. That is the standard location.

The /etc/terminfo is supposedly for letting administrators locally install a custom termcap entry. This is a truly ancient relic from the time you might have wanted to connect a nonstandard serial terminal to your system that was missing from the terminfo database. I haven't seen it used in 3 decades. In fact, you rarely see any terminal other than xterm these days.

The only purpose served by /etc/termcap is to convince people that setting the default terminfo search path is more complicated than it really is.

PR #226, opened 2024-03-06
Upgrade to musl libc to 1.2.5

charliermarsh:
## Summary

Closes https://github.com/indygreg/python-build-standalone/issues/224.

### DIFF
--- a/pythonbuild/downloads.py
+++ b/pythonbuild/downloads.py
@@ -209,10 +209,10 @@ DOWNLOADS = {
         "license_file": "LICENSE.mpdecimal.txt",
     },
     "musl": {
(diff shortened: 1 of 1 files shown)


indygreg (approved):
CI says this _just worked_! I'm pleasantly surprised, as musl has historically caused problems during upgrades. Maybe that's a sign that musl is gaining sufficient traction and stability in the wild. Who knows.

Thanks for the upgrade.

charliermarsh:
I saw 0d64be87e0f32393280ff14ff6f5ecfa2388e4ee and crossed my fingers that it would be straightforward :)

PR #234, opened 2024-03-18
Remove Windows static builds

charliermarsh:
## Summary

This PR removes the Windows static builds from the build and release pipelines. As per https://github.com/indygreg/python-build-standalone/issues/221:

> The Windows static builds in this project are an artifact from trying to achieve the original goals of PyOxidizer. With the benefit of hindsight, the static Windows builds are borderline unusable.

Closes https://github.com/indygreg/python-build-standalone/issues/221.

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -48,8 +48,7 @@ pub static RELEASE_TRIPLES: Lazy<BTreeMap<&'static str, TripleRelease>> = Lazy::
 
     // Windows.
 
-    // The -shared and -static parts of the triple are a lie. But the code
-    // handles it fine.
+    // The -shared part of the triple is a lie. But the code handles it fine.
     h.insert(
         "i686-pc-windows-msvc-shared",
         TripleRelease {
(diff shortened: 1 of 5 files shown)


zanieb (approved):

src/release.rs
```diff
@@ -48,8 +48,7 @@ pub static RELEASE_TRIPLES: Lazy<BTreeMap<&'static str, TripleRelease>> = Lazy::
 
     // Windows.
 
-    // The -shared and -static parts of the triple are a lie. But the code
-    // handles it fine.
+    // The -shared part of the triple is a lie. But the code handles it fine.
```
indygreg, line 51: As a potential future followup we can strip the `-shared` part from the triple. The only reason it is there is to disambiguate from the static builds. Now since there are only dynamically linked builds, there's no need for the label.

I'll file an issue to track this.

indygreg (approved):
As I reviewed all the deleted code, I couldn't help but reflect on how many hours it took to devise those lines of code. Dozens. If not over 100.

It was a bold experiment. But sometimes you can only learn through failure.

Thank you for this PR: it looks great!

src/release.rs
charliermarsh: Cool, I can do that too.

charliermarsh:
> As I reviewed all the deleted code, I couldn't help but reflect on how many hours it took to devise those lines of code. Dozens. If not over 100.

Honestly... I guessed as much. There were so many little details here in the static library conversion, `remove_from_extension_modules, etc. Props for making it possible.

PR #235, opened 2024-03-18
Upgrade armv7 builds to release

charliermarsh:
## Summary

Upgrades `armv7-unknown-linux-gnueabi` and the hard-float variant `armv7-unknown-linux-gnueabihf` to be included in the release. These variants are already built, though they haven't been extensively tested as per the linked issue.

Closes https://github.com/indygreg/python-build-standalone/issues/203.

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -115,6 +115,24 @@ pub static RELEASE_TRIPLES: Lazy<BTreeMap<&'static str, TripleRelease>> = Lazy::
     );
 
     h.insert(
+        "armv7-unknown-linux-gnueabi",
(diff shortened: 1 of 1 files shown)


indygreg (approved):
Nice!

indygreg:
> how would you suggest testing these?

https://gcc.gnu.org/wiki/CompileFarm is always a good bet for access to random machine architectures.

In the past, I've published releases from new architectures without any explicit testing. I ask a potential user to spot test them. Or I just call out the builds are untested in the release notes and rely on people to file issues if anything is seriously broken.

PR #241, opened 2024-03-30
Remove `shared-` prefix from Windows profiles

charliermarsh:
## Summary

In https://github.com/indygreg/python-build-standalone/pull/234, we removed the Windows static builds from python-build-standalone. As such, the `shared-` prefix (in, e.g., `shared-pgo`) is no longer necessary, since there's no need to differentiate the shared builds from the static builds.

This PR removes the `shared-` prefix, but continues to publish under both profiles to allow for a transition period for consumers of the `shared-` builds.

Closes https://github.com/indygreg/python-build-standalone/issues/237.

## Test Plan

Downloaded one of the release artifacts locally:

```
~/Downloads on  main [!+?] via 🐍 v3.12.0
❯ cd cpython-3.8-vcvars64.bat-pgo
~/Downloads/cpython-3.8-vcvars64.bat-pgo on  main [!+?]
❯ ls
cpython-3.8.18-x86_64-pc-windows-msvc-pgo-20240330T1749.tar.zst
cpython-3.8.18-x86_64-pc-windows-msvc-shared-pgo-20240330T1749.tar.zst
~/Downloads/cpython-3.8-vcvars64.bat-pgo on  main [!+?]
❯ shasum cpython-3.8.18-x86_64-pc-windows-msvc-*
718227dbe11dcdf74283952a6b01a769cd5e2919  cpython-3.8.18-x86_64-pc-windows-msvc-pgo-20240330T1749.tar.zst
718227dbe11dcdf74283952a6b01a769cd5e2919  cpython-3.8.18-x86_64-pc-windows-msvc-shared-pgo-20240330T1749.tar.zst
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -51,11 +51,7 @@ jobs:
         vcvars:
           - 'vcvars32.bat'
           - 'vcvars64.bat'
-        # The 'shared-' prefix is no longer needed (i.e., both 'shared-pgo' and 'pgo' resolve to the
-        # same profile). However, we're double-publishing under both names during the transition
-        # period.
         profile:
-          - 'shared-pgo'
           - 'pgo'
     needs: pythonbuild
     runs-on: 'windows-2019'
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -51,8 +51,12 @@ jobs:
         vcvars:
           - 'vcvars32.bat'
           - 'vcvars64.bat'
+        # The 'shared-' prefix is no longer needed (i.e., both 'shared-pgo' and 'pgo' resolve to the
+        # same profile). However, we're double-publishing under both names during the transition
+        # period.
         profile:
           - 'shared-pgo'
+          - 'pgo'
     needs: pythonbuild
     runs-on: 'windows-2019'
     steps:
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -1384,8 +1384,8 @@ def build_cpython(
     openssl_archive,
     libffi_archive,
     openssl_entry: str,
-):
-    pgo = profile in ("pgo", "shared-pgo")
+) -> pathlib.Path:
+    pgo = profile == "pgo"
 
     msbuild = find_msbuild(msvc_version)
     log("found MSBuild at %s" % msbuild)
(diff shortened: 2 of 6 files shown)


.github/workflows/windows.yml
```diff
@@ -51,8 +51,12 @@ jobs:
         vcvars:
           - 'vcvars32.bat'
           - 'vcvars64.bat'
+        # The 'shared-' prefix is no longer needed (i.e., both 'shared-pgo' and 'pgo' resolve to the
+        # same profile). However, we're double-publishing under both names during the transition
+        # period.
         profile:
           - 'shared-pgo'
+          - 'pgo'
```
charliermarsh, line 59: I opted to just run the build twice (even though it will produce the same artifact). I considered running the build once, then copying over the artifact, but candidly I wasn't sure if that would break other assumptions in the pipeline. For example, this job publishes under the name `name: ${{ matrix.py }}-${{ matrix.vcvars }}-${{ matrix.profile }}`, so that would now be "incorrect" for the shared profiles.

cpython-windows/build.py
```diff
@@ -894,12 +893,11 @@ def build_openssl_for_arch(
     # uplink.c tries to find the OPENSSL_Applink function exported from the current
     # executable. However, it is exported from _ssl[_d].pyd in shared builds. So
     # update its sounce to look for it from there.
-    if "shared" in profile:
```
charliermarsh, line 897: (Always true. Should've been removed in the preceding PR.)

.github/workflows/windows.yml
indygreg: A problem with this is PGO is non-deterministic. So the builds won't be bit-fit-bit identical.

Having lived through non-deterministic builds in Firefox due to bugs in MSVC's PGO implementation, I am sensitive to having avoidable build variance.

I would strongly prefer to publish the same binary asset under different names.

Let's change the official profile to `pgo` in the building interface. But either change the release publishing code to upload a 2nd copy. Or change the building to emit multiple `.tar.zst` artifacts. I _think_ the release automation will _just work_ if there are multiple `.tar.zst` in the same GitHub Actions artifact. This is how we used to do it until recently.

indygreg:
I manually cancelled actions for Linux and macOS due to xz download bustage. Rebase on `main` to get working CI.

.github/workflows/windows.yml
charliermarsh: Sounds good, can do. I think the latter will be more straightforward (change `cpython-windows/build.py` to emit two artifacts).

charliermarsh: (This makes a lot of sense.)

indygreg (approved):
This looks like it should work! Thanks for following up!

PR #245, opened 2024-04-01
Add automation for lint, format, and type check

zanieb:
Follows https://github.com/indygreg/python-build-standalone/pull/236#discussion_r1529586740

Adds a `check.py` script and splits some requirements into `requirements.dev.in` / `.txt`. Enforces checks in a CI job.

Runs:

- `ruff check` with a very limited selection of rules, can add more later
- `ruff format`
- `mypy` on a subset of modules

A `--fix` flag can be provided to apply linter fixes and reformat files.


indygreg (approved):
This was long overdue. Thank you for doing this!

PR #249, opened 2024-04-13
Bump to CPython 3.11.9, 3.12.3

charliermarsh:


indygreg (approved):
Nice.

Need to wait for GitHub CI to clear the builds. But these patch releases usually _just work_. Will merge after I see enough green.

Issue #244, opened 2024-04-01
Release for Python 3.8.19, 3.9.19, and 3.10.14


indygreg:
CPython 3.8.19, 3.9.19, and 3.10.14 were released on March 19, 2024 and contain security fixes. We should upgrade and get a release out the door with these versions.

zanieb:
I'd be interested in doing some work on this.

charliermarsh:
I'll do the download bump now to start.

indygreg:
@charliermarsh you are a collaborator on the repo now and the latest commit on `main` (`d09ff921d92d6da8d8a608eaa850dc8c0f638194`) should be appropriate for a release (assuming CI passes). You should be able to run `just release $(cat /path/to/github/token) d09ff921d92d6da8d8a608eaa850dc8c0f638194 20240414` to perform a release.

You have my blessing to run this command.

The release should be created as pre-release by default. Please make an attempt to populate the release notes. Then @ me here once done so I can look at things and we can promote the release.

If you run into release pipeline regressions during the release (not uncommon!), I usually just hack up code locally to get it to pass, push to `main` [via PR], then run the release automation from the fixed commit, targeting an older commit (so you don't have to wait for GitHub Actions to finish before releasing).

charliermarsh:
Great, thanks @indygreg! I'll give this a try today.

charliermarsh:
Is it possibly because the builds are all now re-running on that commit, since we created a tag for it?

indygreg:
The Rust release automation is only supposed to look at CI invocations that are done. So multiple CI runs in the same commit shouldn't be an issue. But this logic has historically been buggy.

Earlier log snippet indicates the CI artifact wasn't a zip file. Feels like an intermittent failure or we messed up artifact uploading.

charliermarsh:
Wondering if there's something wrong with my creds because I'm seeing it consistently for different artifacts; and when I download them out-of-band, they seem to unzip without error (e.g., `cpython-3.11-aarch64-apple-darwin-debug`). Looking into it.

charliermarsh:
(Creds would be odd though since I'd expect an earlier failure _and_ my creds successfully created the Release itself.)

charliermarsh:
Okay yeah the fetched artifact on my machine (when fetched via `github.rs`) is just 158 bytes. Looking at it...

charliermarsh:
Hah, ok, the contents are just:
```json
[...]

PR #250, opened 2024-04-15
Avoid incorrect triple match for substrings

charliermarsh:
## Summary

Given an artifact like `cpython-3.8.19-x86_64-pc-windows-msvc-shared-pgo-20240401T1106.tar.gz`, we were matching against the triple `x86_64-pc-windows-msvc` (rather than `x86_64-pc-windows-msvc-shared`); so the release code then assumed that the install suffix was `shared-pgo` rather than `pgo`, and failed to align the assets.

This should fix _both_ of the issues described in #244, since the hard-float armv7 triples are _also_ prefixes of the non-hard-float variants (`armv7-unknown-linux-gnueabihf` vs. `armv7-unknown-linux-gnueabi`).

## Test Plan

Ran `just release` with these local changes (skipping uploads); verified that no missing-file errors were raised.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -204,7 +204,9 @@ pub async fn command_fetch_release_distributions(args: &ArgMatches) -> Result<()
 
             let name = zf.name().to_string();
 
-            if let Some((triple, release)) = RELEASE_TRIPLES.iter().find_map(|(triple, release)| {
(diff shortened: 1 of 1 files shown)


zanieb (approved):

indygreg (approved):
Good catch!

PR #251, opened 2024-04-17
fix: bad tag for latest-release.json

Yohe-Am:
Hey, looks like something went awry and the `latest-release.json` file now sports a sha instead of the tag. This PR fixes it.

### DIFF
--- a/latest-release.json
+++ b/latest-release.json
@@ -1,6 +1,6 @@
 {
   "version": 1,
-  "tag": "d09ff921d92d6da8d8a608eaa850dc8c0f638194",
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks!

charliermarsh:
Oh I may have messed up by using the SHA somewhere instead of the tag. But I think this should be `20240415`. I'm not sure _why_ there isn't an entry in the history for `20240224`. \cc @indygreg

charliermarsh:
I'm worried that the SHA will break consumers so going to merge even though this points to the "last" release.

PR #257, opened 2024-04-21
Add a release workflow to GitHub Actions

### LINKED ISSUE #253: `just release` should be idempotent
If you run `just release` twice, `gh release create --prerelease --notes TBD --title {{ tag }} --target {{ commit }} {{ tag }}` will fail the second time, since the release already exists. Can we make it robust to repeated invocations with the same data?

### LINKED ISSUE #256: Move release process to GitHub Actions
See: https://github.com/indygreg/python-build-standalone/issues/244#issuecomment-2065676823

charliermarsh:
## Summary

This PR moves the release workflow to GitHub Actions, which removes the need for wrangling PATs and other access controls locally.

Specifically, we now have a "Release" workflow that runs via `workflow_dispatch`, and accepts the commit SHA and the tag as arguments. You can run the workflow on a specific branch, which defaults to `main`, so the commit on which the release workflow _runs_ does not need to match the commit being _released_, which IMO is a nice property.

The workflow is idempotent, such that it's safe to re-run with the same commit and tag if the release has already started, or completed, or failed partway through.

There's also a "dry-run" mode that avoids creating the release or uploading any artifacts. (If the release _does_ exist, it will go a bit "further" in the process.)

Closes #253.
Closes #256.

## Test Plan

This took an embarrassing amount of trial and error, but feels pretty good now.

- Ran to completion: https://github.com/charliermarsh/python-build-standalone/actions/runs/8772527711
- Attempted to run with the same tag but different commit, which (correctly) raised an error: https://github.com/charliermarsh/python-build-standalone/actions/runs/8772541686
- Re-ran with same tag same commit, which correctly validated but didn't change anything, and exited successfully: https://github.com/charliermarsh/python-build-standalone/actions/runs/8772581636
- Ran with dry-run, which didn't create a release but validated the artifacfts: https://github.com/charliermarsh/python-build-standalone/actions/runs/8772642509

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,74 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "The version to release (e.g., '20240414')."
+        type: string
+      sha:
+        description: "The full SHA of the commit to be released (e.g., 'd09ff921d92d6da8d8a608eaa850dc8c0f638194')."
+        type: string
+      dry-run:
+        description: "Whether to run the release process without actually releasing."
+        default: false
+        required: false
+        type: boolean
+
+permissions:
+  contents: write
+  packages: write
+
+jobs:
+  release:
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: extractions/setup-just@v2
+
+      # Perform a release in dry-run mode.
+      - run: just release-dry-run ${{ secrets.GITHUB_TOKEN }} ${{ github.event.inputs.sha }} ${{ github.event.inputs.tag }}
+        if: ${{ github.event.inputs.dry-run == 'true' }}
+
+      # Create the release itself.
+      - name: Configure Git identity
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Create GitHub Release
+        if: ${{ github.event.inputs.dry-run == 'false' }}
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ github.event.inputs.tag }}
+          name: ${{ github.event.inputs.tag }}
+          prerelease: true
+          body: TBD
+          allowUpdates: true
+          updateOnlyUnreleased: true
```
charliermarsh, line 70: We can do this with `gh`, but the action is helpful because of these `allowUpdates` and `updateOnlyUnreleased` properties.

This action is also used in `cargo-dist` (and therefore we use it in `ruff` and `uv`), so feels well-trodden.

src/github.rs
```diff
@@ -48,7 +48,7 @@ async fn upload_release_artifact(
     dry_run: bool,
 ) -> Result<()> {
```
charliermarsh, line 49: The code changes in this file are just making `upload_release_artifact` more robust in dry-run, such that if the release itself doesn't exist, it can still do some validation.

charliermarsh:
I'm going to bias towards merging this so that I can use it for an upcoming release.

PR #266, opened 2024-06-04
running.rst: explain that install_only is based on the fastest build configuration

noamraph:
Today I downloaded an archive, after reading the Running Distribution docs, and had a bit of a hard time in deciding which archive I wanted, since the `install_only` archive file name doesn't have the build configuration name, so I wasn't sure which build configuration I got. Later I read the Distribution Archive doc and understood that it includes the fastest configuration. I think it's worthwhile to write this in the Running Distribution docs as well.

Thanks for your hard work!
Noam


indygreg (approved):
Thank you for clarifying this!

PR #269, opened 2024-06-06
fix: detect preview versions of msvs

### LINKED ISSUE #268: Fails to build on Win 11 with Preview visual studio
Command line:
```cmd
py.exe build-windows.py --profile noopt --vs 2022 --python cpython-3.12 --sh c:\cygwin64\bin\bash.exe
```
downloads strawberry perl, process libffi, and then fails with the obtuse: `VC\Auxiliary\Build\vcvarsall.bat not found.`

This appears to be two fold:

1- the result of the subcommand in find_vs_path is not checked for emptiness,
2- the command is not passing the "-prerelease" option to vswhere,

An easy fix would be:

```
diff --git a/cpython-windows/build.py b/cpython-windows/build.py
index bc97cd5..66ff6a1 100644
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -233,18 +233,22 @@ def find_vs_path(path, msvc_version):
     else:
         raise ValueError(f"unsupported Visual Studio version: {msvc_version}")

-    p = subprocess.check_output(
-        [
-            str(vswhere),
-            # Visual Studio 2019.
-            "-version",
-            version,
-            "-property",
-            "installationPath",
-            "-products",
-            "*",
-        ]
-    )
+    cmdline = [
+        str(vswhere),
+        # Visual Studio 2019.
+        "-version",
+        version,
+        "-property",
+        "installationPath",
+        "-prerelease",
+        "-products",
+        "*",
+    ]
+    p = subprocess.check_output(cmdline)
+    if not p:
+        cmd = ' '.join('"%s"' % i for i in cmdline)
+        print("could not find visual studio (using '%s')" % cmd)
+        sys.exit(1)
```

But I suspect this would kick the can down the road to an additional problem if you had vs2022 _and_ vs2022 preview installed.

kfsone:
fixes #268

- add the '-prerelease' flag to the vswhere subcommand,
- give 'no vs found' error if the command returns empty string,
- catch vswhere returning multiple installations,

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -224,8 +224,20 @@ def find_vswhere():
 
 
 def find_vs_path(path, msvc_version):
+    """
+    Attempts to find 'path' by invoking vswhere to identify where the
+    given msvc_version is installed.
+
+    vswhere is akin to things like 'pkg-config', it's an executable
+    that can be invoked to retrieve parameters of visual studio code
+    installations on the host machine.
+    """
     vswhere = find_vswhere()
 
+    # vswhere uses a math-like range expression for specifying the
+    # major release numbers.
+    #  16 = Visual Studio 2019,
+    #  17 = Visual Studio 2022,
     if msvc_version == "2019":
         version = "[16,17)"
     elif msvc_version == "2022":
@@ -233,26 +245,42 @@ def find_vs_path(path, msvc_version):
     else:
         raise ValueError(f"unsupported Visual Studio version: {msvc_version}")
 
-    p = subprocess.check_output(
-        [
-            str(vswhere),
-            # Visual Studio 2019.
-            "-version",
-            version,
-            "-property",
-            "installationPath",
-            "-products",
-            "*",
-        ]
-    )
+    # Assemble the command line into an argv list for a subprocess
+    # call, so if the command fails we can tell the user what command
+    # we tried and enable them to potentially investigate further.
+    cmdline = [
+        str(vswhere),
+        "-version",
+        version,
+        "-property",
+        "installationPath",
+        "-prerelease",
+        "-products",
+        "*",
+    ]
+    p = subprocess.check_output(cmdline)
+    if not p:
+        # add quotes around any of the terms with spaces in them, so that
+        # the user can copy-and-paste the command.
+        cmdline = ['"%s"' % term if " " in term else term for term in cmdline]
+        print("could not find visual studio (using '%s')" % cmdline)
+        sys.exit(1)
 
     # Strictly speaking the output may not be UTF-8.
-    p = pathlib.Path(p.strip().decode("utf-8"))
+    p = p.strip().decode("utf-8")
 
-    p = p / path
+    # vswhere can potentially report multiple installed versions,
+    # one per line. The above 'strip' will have removed any
+    # trailing newlines, so if there is still a newline, there
+    # is more than one option.
+    if "\n" in p:  # more than one line
+        print("found multiple matching instances of visual studio %s:" % msvc_version)
+        print(p)
+        sys.exit(1)
 
+    p = pathlib.Path(p, path)
     if not p.exists():
-        print("%s does not exist" % p)
+        print("%s does not exist for vs%s" % (p, msvc_version))
         sys.exit(1)
 
     return p


cpython-windows/build.py
```diff
@@ -233,26 +237,34 @@ def find_vs_path(path, msvc_version):
...
+        # Visual Studio 2019.
+        "-version",
+        version,
+        "-property",
+        "installationPath",
+        "-prerelease",
+        "-products",
+        "*",
+    ]
+    p = subprocess.check_output(cmdline)
```
zanieb, line 251: Nit: Could you avoid this change if it's not necessary? It makes the diff harder to read and it just seems stylistic.

cpython-windows/build.py
```diff
@@ -233,26 +237,34 @@ def find_vs_path(path, msvc_version):
...
-    # Strictly speaking the output may not be UTF-8.
-    p = pathlib.Path(p.strip().decode("utf-8"))
+    p = p.strip().decode("utf-8")
+    if "\n" in p:  # more than one line
+        print("found multiple matching instances of visual studio %s:" % msvc_version)
+        print(p)
+        sys.exit(1)
 
-    p = p / path
+    # Strictly speaking the output may not be UTF-8.
```
zanieb, line 263: Should this comment remain up where we decode the output?

cpython-windows/build.py
zanieb: Ah I see you use it downstream in a new error message. That seems okay.

cpython-windows/build.py
```diff
@@ -233,26 +237,34 @@ def find_vs_path(path, msvc_version):
...
+        version,
+        "-property",
+        "installationPath",
+        "-prerelease",
+        "-products",
+        "*",
+    ]
+    p = subprocess.check_output(cmdline)
+    if not p:
+        cmd = " ".join('"%s"' % i for i in cmdline)
```
zanieb, line 253: Is there a reason to quote each item in the invocation? I feel like it'd probably be more readable without.

zanieb:
Thanks for contributing!

zanieb:
Is there a compelling reason to support building with prerelease versions of Visual Studio?

cpython-windows/build.py
```diff
@@ -233,26 +237,34 @@ def find_vs_path(path, msvc_version):
...
+        print("could not find visual studio (using '%s')" % cmd)
+        sys.exit(1)
 
-    # Strictly speaking the output may not be UTF-8.
-    p = pathlib.Path(p.strip().decode("utf-8"))
+    p = p.strip().decode("utf-8")
+    if "\n" in p:  # more than one line
+        print("found multiple matching instances of visual studio %s:" % msvc_version)
```
charliermarsh, line 259: So this is like: you have 2022 and the 2022 pre-release installed? Should we just... pick one? (If you know) are they ordered?

cpython-windows/build.py
kfsone: It's there so that it can tell the user what command-line it tried to run; don't mind reverting it but my past experience trying to wrangle vcvars into cooperation was that people tended to ask "well what command did you run?" :)

cpython-windows/build.py
kfsone: I was trying to balance between copy-pasteability and complexity. The first argument is typically going to have multiple spaces in it :/

"C:\Program Files (x86)\Microsoft\Visual Studio 2019\".. le sigh :( Quoting only the ones that have spaces in is doable but kind of hard to read:

```
cmd = " ".join([
  '"%s"' if ' ' in i else i
  for i in cmdline
  ])

**Comment by kfsone** [comment]
> Is there a compelling reason to support building with prerelease versions of Visual Studio?

Ah - I can see why you might wonder that; "prerelease" recognizes the "Preview" aka canary editions. Which is not the same as whatever alpha or beta of 2024 or 2025 might eventually come into existence.

This is a fairly common pattern:

image

Stable + Firehose.

File: cpython-windows/build.py (hunk 4 above)
Comment (kfsone) [reply cpython-windows/build.py:259]: I was a little surprised to find someone else fighting a similar fight with vs version stuff. The studio I work for has a python-based wrapper (generate.py) around cmake for buffering our non-tech employees from having to learn how to drive cmake for all our various SKUs and environments, and I'd had to build a bunch of stuff into that dealing with vswhere and vcvars.

Unfortunately, about 14 months ago I found a complete workaround: switching to powershell (*) and using the Visual Studio DevShell module's Enter-VsDevShell command.

I never figured out what the difference was between _that_ and just opening a developer cmd prompt was. But with the cmd version, the cmake our wrapper invokes can't find a visual studio and I end up with a bunch of python to find it by hand and inject the variables into the env; using the ps approach just works.

But - I was fighting _all_ the fires at the time, including a Python 2 -> 3 migration, at the start of the work so it's possible, probable?, I just missed something really trivial.
 
(* we were resurrecting the Windows build of 15+ year old code that had spent 8 years mostly on Mac; Open Source + Cross-Platform powershell had just come out; I didn't know squat about Jenkins and I had 30 years experience with unix shells and that-time-I-played-jedi-knight total experience with dos. I figured using PS as a rosetta stone would help me find my feet or give me great material for a blong-rant... It turned out to be like going from Perl to Python. I actively switch between mac/win/lin these days, because I can take my scripts with me without any resistance)

Comment (kfsone) [reply cpython-windows/build.py:259]: For context, my current machine reports this if I remove the `-version` argument:

```
E:\>"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -property installationPath -prerelease -products *
C:\Program Files\Microsoft Visual Studio\2022\Preview
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools
```

I definitely installed Vs2022 preview after BuildTools, but I did just updated Vs2022 preview, and I couldn't find documentation on the ordering.

File: cpython-windows/build.py (hunk 2 above)
Comment (kfsone) [reply cpython-windows/build.py]: Yep, fixed in latest commit along with some additional comments.

File: cpython-windows/build.py (hunk 4 above)
Comment (kfsone) [reply cpython-windows/build.py:275]: There are two options:
```
  -latest        Return only the newest version and last installed.
  -sort          Sorts the instances from newest version and last installed to oldest.
                 When used with "find", first instances are sorted then files are sorted lexigraphically.
```

outputs:

```
E:\>"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -property installationPath -prerelease -products *
C:\Program Files\Microsoft Visual Studio\2022\Preview
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools

E:\>"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -property installationPath -prerelease -products * -sort
C:\Program Files\Microsoft Visual Studio\2022\Preview
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools

E:\>"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -property installationPath -prerelease -products * -latest
C:\Program Files\Microsoft Visual Studio\2022\Preview
```

ergo we could use '-latest', but that's not really solving the problem.

It would be nice if running this from a devshell could just use the _current_ visual studio, but I lost that fight several times trying to do something similar with our studio's python-based cmake wrapper and haven't had the courage to revisit it since finding using powershell was a sufficient workaround =/

charliermarsh: Sorry, can you clarify why using `-latest` isn't really solving the problem?

kfsone: If I've been building time-and-again with C++X language features only available in the Preview edition and then it suddenly fails on me because it transparently switched to using the latest update of the Build Tools release, that's going to be a head scratcher, especially if it was an automatic update nobody manually triggered.

If they have multiple installs of the same edition, I'm sure they're expecting to have a switch of some kind for it, but I don't really want to explode *this* diff into tackling that.

It's not uncommon for canary builders or ci machines to have multiple VS variants of the same edition installed: 1/ VS 2022 Build Tools, 2/ VS 2022 Professional, 3/ VS 2022 Ultimate Preview, or if they have 4/ a Community Edition installed that gives them licensing considerations.

cpython-windows/build.py
Edward-Knight: [`shlex.join`](https://docs.python.org/3/library/shlex.html#shlex.join) may be what you're looking for. The quoting isn't designed for non-POSIX shells but should do the job for pretty-printing.

PR #270, opened 2024-06-06
Fix clippy lints

zanieb:
Updates the code to pass `cargo clippy`. Most of the fixes were auto-applied by Clippy.

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -792,7 +792,7 @@ impl ValidationContext {
     }
 }
 
-fn validate_elf<'data, Elf: FileHeader<Endian = Endianness>>(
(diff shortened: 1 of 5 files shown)


src/validation.rs
```diff
@@ -1904,6 +1899,7 @@ fn validate_distribution(
 
             let exported = context.libpython_exported_symbols.contains(&ext.init_fn);
 
+            #[allow(clippy::needless_bool, clippy::if_same_then_else)]
```
zanieb, line 1902: I opted to allow the if statements to keep their current structure though it is redundant because I find it quite clear this way and am not compelled to change the style.

charliermarsh (approved):
Thanks!

PR #301, opened 2024-08-24
Stop bundling setuptools for Python 3.12+

### LINKED ISSUE #300: Stop bundling setuptools for Python 3.12+
In Python 3.12, the stdlib's `ensurepip` and `venv` modules were updated to no longer install `setuptools` alongside pip. This was viable since as of pip v22.1, for pre-PEP-517/518 packages [pip will now](https://github.com/pypa/pip/pull/10717) default to the isolated build environment mode (along with a fallback legacy setuptools build backend, with setuptools and wheel automatically installed), if the setuptools package isn't installed globally.

As such, when using the default Python configure options (which include `--with-ensurepip`, which causes ensurepip to run during the build), the resultant Python build only includes pip and not setuptools for Python 3.12+.

For example, this then means the Python distribution used by the official GitHub Actions [setup-python](https://github.com/actions/setup-python) action (which don't override the upstream CPython configure defaults) automatically only ships setuptools for Python 3.11 and older:

```
# Python 3.11
$ curl -sSfL https://github.com/actions/python-versions/releases/download/3.11.9-9947079978/python-3.11.9-linux-24.04-arm64.tar.gz | tar -zt | grep -E 'site-packages/[^/]+/$'
./lib/python3.11/site-packages/pip-24.0.dist-info/
./lib/python3.11/site-packages/_distutils_hack/
./lib/python3.11/site-packages/pkg_resources/
./lib/python3.11/site-packages/pip/
./lib/python3.11/site-packages/setuptools/
./lib/python3.11/site-packages/setuptools-65.5.0.dist-info/
```

```
# Python 3.12
$ curl -sSfL https://github.com/actions/python-versions/releases/download/3.12.5-10375840348/python-3.12.5-linux-24.04-arm64.tar.gz | tar -zt | grep -E 'site-packages/[^/]+/$'
./lib/python3.12/site-packages/pip-24.2.dist-info/
./lib/python3.12/site-packages/pip/
```

However, the builds performed by this project configure Python using `--without-ensurepip`, and then manually install pip and setuptools afterwards:
https://github.com/indygreg/python-build-standalone/blob/90d40b46d929eabafee1e3320ab54d32efdf258e/cpython-unix/build-cpython.sh#L546-L547

This means that when Python 3.12 builds were added in 2a5f3a9d4a4c1b15c2bf8db26a042d1d1efb29d6 they didn't automatically pick up this upstream CPython change.

For parity with upstream Python's behaviour, the builds here should stop explicitly installing setuptools for Python 3.12+.

See also:
- https://github.com/python/cpython/issues/95299
- https://github.com/python/cpython/commit/ece20dba120a1a4745721c49f8d7389d4b1ee2a7
- https://github.com/pypa/virtualenv/issues/2487
- https://github.com/astral-sh/uv/pull/1613
- https://github.com/pypa/get-pip/pull/218
- https://github.com/docker-library/python/pull/954

edmorley:
For parity with the `ensurepip` and `venv` modules, which no longer install setuptools as of Python 3.12:
- https://github.com/python/cpython/issues/95299
- https://github.com/python/cpython/commit/ece20dba120a1a4745721c49f8d7389d4b1ee2a7

Fixes #300.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -576,7 +576,12 @@ fi
 # simply use our pip to install self. Kinda crazy, but it works!
 
 ${BUILD_PYTHON} "${PIP_WHEEL}/pip" install --prefix="${ROOT}/out/python/install" --no-cache-dir --no-index "${PIP_WHEEL}"
-${BUILD_PYTHON} "${PIP_WHEEL}/pip" install --prefix="${ROOT}/out/python/install" --no-cache-dir --no-index "${SETUPTOOLS_WHEEL}"
+
+# Setuptools is only installed for Python 3.11 and older, for parity with
+# `ensurepip` and `venv`: https://github.com/python/cpython/pull/101039
+if [ -n "${PYTHON_MEETS_MAXIMUM_VERSION_3_11}" ]; then
+    ${BUILD_PYTHON} "${PIP_WHEEL}/pip" install --prefix="${ROOT}/out/python/install" --no-cache-dir --no-index "${SETUPTOOLS_WHEEL}"
+fi
 
 # Hack up the system configuration settings to aid portability.
 #
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -19,6 +19,7 @@ import zipfile
 
 from pythonbuild.cpython import (
     STDLIB_TEST_PACKAGES,
+    meets_python_maximum_version,
     meets_python_minimum_version,
     parse_config_c,
 )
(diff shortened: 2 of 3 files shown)


indygreg:
This PR seems technically sound. But I'd like a few questions answered at https://github.com/indygreg/python-build-standalone/issues/300#issuecomment-2308528284 before merging this.

indygreg (approved):

edmorley:
I've rebased on `main` after the recent PR merges.

edmorley:
Is this ready to be merged? :-)

zanieb:
Thanks for rebasing! I'll review this soon. I might ship this after the other changes.

zanieb (approved):

zanieb:
Eh, after thinking a bit more I don't really want to ship `setuptools` with Python 3.13 if we're going to remove it anyway, so it seems better to do it all at once.

edmorley:
Many thanks :-)

PR #308, opened 2024-08-27
Disable _dbm extension module

d3dave:
Disable _dbm extension module everywhere and remove all usages of BerkeleyDB which it depends on.

Also, adjust the Linux workflow to build only 3.9+, non-debug, arm64+x86_64 targets.


d3dave:
:sigh: wrong target branch 😅

PR #312, opened 2024-09-01
Ensure that Hacl_Hash_SHA2 does not appear as a library to link to

ntamas:
When trying to update PyOxidizer and PyOxy to support standalone builds of Python 3.12, I noticed that the build attempted to link to a library named `Hacl_Hash_SHA2` when `pyoxy` was linked (with `-lHacl_Hash_SHA2`). This PR fixes the metadata in `PYTHON.json`.

### DIFF
--- a/cpython-unix/extension-modules.yml
+++ b/cpython-unix/extension-modules.yml
@@ -411,17 +411,12 @@ _sha2:
   minimum-python-version: "3.12"
   sources:
     - sha2module.c
+    - _hacl/Hacl_Hash_SHA2.c
   includes:
     - Modules/_hacl/include
-    - Modules/_hacl/internal
   defines:
     - _BSD_SOURCE
     - _DEFAULT_SOURCE
-  links:
-    # Use the colon syntax to prevent the library dependency from getting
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Makes sense to me if the builds pass.

zanieb:
Can you share a diff of the `PYTHON.json` metadata?

ntamas:
Sure! I'm attaching two diffs here; `full.diff` was generated with `diff -u`, but it contains lots of lines where the only difference is in the names of the temporary folders where the build was executed. `relevant.diff` was created manually from `full.diff` by deleting those chunks where the only difference was the name of the temporary folder so you can see the differences more clearly.

Both diffs are attached in a ZIP file here: [diffs.zip](https://github.com/user-attachments/files/16905952/diffs.zip)

zanieb:
Thanks! The change in the metadata makes sense to me but I don't fully understand why it was written this way in the first place?

ntamas:
I don't know either -- maybe @indygreg remembers? I think he was the one who initially wrote the code for supporting Python 3.12 where the pre-3.12 hash functions were replaced with implementations from HACL*.

PR #313, opened 2024-09-04
tcl/tk: upgrade from 8.6.12 -> 8.6.14 on Windows

zanieb:
Attempt to work through some of the failures in https://github.com/indygreg/python-build-standalone/pull/303

> It seems we have to bring a copy of zlib as a DLL now, so the installers will need an update. I'm also a little bit concerned about it interfering with other modules, but there doesn't seem to be an easy way to statically link it anymore, so I guess we'll just have to deal with any fallout if it breaks stuff.

from https://github.com/python/cpython/issues/99834#issuecomment-1329822722


zanieb:
Failing on Windows still with

```
cpython>          C:\Users\runneradmin\AppData\Local\Temp\python-build-7ct6gfb1\Python-3.8.19\Modules\_tkinter.c(47,10): fatal error C1083: Cannot open include file: 'tcl.h': No such file or directory [C:\Users\RUNNER~1\AppData\Local\Temp\python-build-7ct6gfb1\Python-3.8.19\PCbuild\_tkinter.vcxproj]
cpython>          C:\Users\runneradmin\AppData\Local\Temp\python-build-7ct6gfb1\Python-3.8.19\Modules\tkappinit.c(16,10): fatal error C1083: Cannot open include file: 'tcl.h': No such file or directory [C:\Users\RUNNER~1\AppData\Local\Temp\python-build-7ct6gfb1\Python-3.8.19\PCbuild\_tkinter.vcxproj]
```

zanieb:
With https://github.com/indygreg/python-build-standalone/pull/313/commits/0bd2a31dc3f7d856f06ad6ffdfe4755b5fb3b637 we're past that and now failing validation checks with

```
  error: python/install/DLLs/tcl86t.dll loads illegal library zlib1.dll
  error: python/install/tcl/nmake/x86_64-w64-mingw32-nmakehlp.exe loads illegal library msvcrt.dll
```

I was hoping the first wouldn't be present in 3.12 due to https://github.com/zanieb/python-build-standalone/blob/zb/tcl-tk-8614/cpython-windows/build.py#L560-L567 and we could just make that patch unconditionally — but it fails on 3.12 with the same error so that doesn't appear relevant.

The latter file was added in https://core.tcl-lang.org/tcltls/info/b7b0bd5a8f8b09e4

src/validation.rs
```diff
@@ -349,6 +349,11 @@ static DARWIN_ALLOWED_DYLIBS: Lazy<Vec<MachOAllowedDylib>> = Lazy::new(|| {
                 max_compatibility_version: "1.0.0".try_into().unwrap(),
                 required: true,
             },
+            MachOAllowedDylib {
+                name: "/System/Library/Frameworks/UniformTypeIdentifiers.framework/Versions/A/UniformTypeIdentifiers".to_string(),
+                max_compatibility_version: "1.0.0".try_into().unwrap(),
```
zanieb, line 354: I don't know how to determine this version.

zanieb:
@indygreg I'm not entirely sure what to do about the zlib situation on Windows. I'll continue to poke at it, but if you have any suggestions I'd appreciate them. I think I'll probably spin up a Windows machine to debug, it's a bit of a struggle via CI.

src/validation.rs
indygreg: Run `otool -L` on the binary requiring this library. This prints the compatibility version that is embedded in the Mach-O library dependency data.

cpython-windows/build.py
```diff
@@ -421,6 +421,13 @@ def hack_props(
             rb"<tcltkDir>%s\$(ArchName)\</tcltkDir>" % tcltk_path,
         )
 
+    try:
+        static_replace_in_file(
+            tcltkprops_path, rb"<tclZlibDLLName>zlib1.dll</tclZlibDLLName>", rb""
+        )
+    except NoSearchStringError:
+        pass
```
indygreg, line 429: This part is slightly concerning.

Does CPython's build of tcl require a `zlib1.dll`? If so, removing this reference may mean we're not shipping a `zlib1.dll` requiring dependency file?

I don't think we have any references to a `zlib1.dll` in this project. If there is a `zlib1.dll` dependency here, that would be concerning.

zanieb: I don't think this patch (https://github.com/indygreg/python-build-standalone/pull/313/commits/97b944863bac338eff82c4b8e61a3bb8ac1a7fa0) actually accomplished anything. I'm just poking around here.

The upstream pull request upgrading the version (https://github.com/python/cpython/pull/101307) says:

> I don't see an easy way to avoid having a second copy of zlib specifically for Tcl, so I guess we just have to include it, but I'm pretty sure that's all handled here.

I'm not sure how to make sure this is linked to our zlib. The Windows distribution validation is failing with with:

```
error: python/install/DLLs/tcl86t.dll loads illegal library zlib1.dll
error: python/install/tcl/nmake/x86_64-w64-mingw32-nmakehlp.exe loads illegal library msvcrt.dll
```

cpython-unix/build.py
```diff
@@ -859,10 +859,10 @@ def build_cpython(
 
         python_info["tcl_library_path"] = "install/lib"
         python_info["tcl_library_paths"] = [
-            "itcl4.2.2",
+            "itcl4.2.4",
             "tcl8",
             "tcl8.6",
-            "thread2.8.7",
+            "thread2.8.79",
```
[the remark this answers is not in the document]
zanieb: Indeed, thanks. I'll double check that.

jjhelmus (approved):

jjhelmus (changes requested):
The project files for _tkinter prior to CPython 3.12 do not include the zlib DLL as an output. This results in this file not being included in the distribution, breaking _tkinter.

Additionally the logic around lines 665 of cpython-windows/build.py removes this file from 3.12 and 3.13 distributions:

``` Python
    # Python 3.12+ uses the the pre-built tk-windows-bin 8.6.12 which doesn't
    # have a standalone zlib DLL, so we remove references to it. For Python
    # 3.14+, we're using tk-windows-bin 8.6.14 which includes a prebuilt zlib
    # DLL, so we skip this patch there.
    # On arm64, we use the new version of tk-windows-bin for all versions.
    if meets_python_minimum_version(python_version, "3.12") and (
        meets_python_maximum_version(python_version, "3.13") or arch == "arm64"
    ):
        try:
            static_replace_in_file(
                pcbuild_path / "_tkinter.vcxproj",
                rb'<_TclTkDLL Include="$(tcltkdir)\bin\$(tclZlibDllName)" />',
                rb"",
            )
        except NoSearchStringError:
            pass
   ```
   
   This file will need to be added to the distribution, either manually or by patching the project files.

PR #279, opened 2024-07-14
Add `install_only_stripped` binaries to release

charliermarsh:
## Summary

This PR adds an `install_only_stripped` variant, which is generated by taking the `install_only` variant and removing debug symbols.

Closes https://github.com/indygreg/python-build-standalone/issues/277.
Closes https://github.com/indygreg/python-build-standalone/issues/174.
Related to https://github.com/indygreg/python-build-standalone/issues/275.

## Test Plan

On macOS:

- Downloaded [cpython-3.10.14+20240713-aarch64-apple-darwin-install_only.tar.gz](https://github.com/indygreg/python-build-standalone/releases/download/20240713/cpython-3.10.14+20240713-aarch64-apple-darwin-install_only.tar.gz) locally.
- Ran: `cargo run convert-install-only-stripped cpython-3.10.14+20240713-aarch64-apple-darwin-install_only.tar.gz`.
- Relocated `cpython-3.10.14+20240713-aarch64-apple-darwin-install_only_stripped.tar.gz` to another directory.
- Unzipped `cpython-3.10.14+20240713-aarch64-apple-darwin-install_only.tar.gz`.
- Ran `./python` in `python/python/bin`.

Performed the same procedure on Windows.

PR #272, opened 2024-06-07
Update to CPython 3.12.4.

kpfleming:

PR #317, opened 2024-09-05
zb/ci cache

zanieb:
- **Disable builds for branches that are not pull requests or `main`**
- **Fix clippy lints (#270)**
- **Update to CPython 3.12.4. (#272)**
- **running.rst: explain that install_only is based on the fastest build configuration**
- **Add a release workflow to GitHub Actions (#257)**
- **Upgrade LLVM 17.0.6 -> 18.0.8**
- **downloads: OpenSSL 3.0.13 -> 3.0.14**
- **downloads: SQLite 3.45.1 -> 3.46.0**
- **downloads: setuptools 69.1.0 -> 70.3.0**
- **downloads: pip 24.0 -> 24.1.2**
- **downloads: ncurses 6.4 -> 6.5**
- **downloads: libffi 3.4.4 -> 3.4.6**
- **Add `install_only_stripped` binaries to release (#279)**
- **Cancel in-flight builds on PR update (#282)**
- **Add Ruff and mypy configuration files (#283)**
- **Create build directory when downloading LLVM (#285)**
- **Only cancel in-progress runs on PRs (#286)**
- **Fix install_only_stripped name translation (#288)**
- ** Avoid overwriting install_only builds in release (#291)**
- **Fix crate build son Rust 1.80.0 (#292)**
- **Avoid duplicating .tar.gz suffix on stripped builds (#293)**
- **Update to CPython 3.12.5 (#299)**
- **license: update LICENSE file to MPL 2.0**
- **Use `Swatinem/rust-cache` instead of `actions/cache`**


charliermarsh (approved):

reneleonhardt:
@charliermarsh Approved but not merged, are there any problems?

zanieb:
It's just not a critical change, why do you ask..?

reneleonhardt:
It has been approved 8 months ago, makes you wonder why not just merge it since then... now it has conflicts.

indygreg:
The Rust cache has marginal improvements to CI times. We can likely improve cache hit times by modifying `actions/cache`.

Given the criticality of PBS to the larger software ecosystem and that popularity making it an attractive target for software supply chain attacks, I have a soft preference for avoiding 3rd party GitHub Actions. If we do adopt 3rd party actions, I think we should pin the Git commit SHA so the implementation can't be changed out from under us without an explicit Git commit to this project.

zanieb:
Agreed, https://github.com/astral-sh/uv/pull/12189 has been working well for us.

PR #318, opened 2024-09-05
Remove custom Rust toolchain installation

zanieb:
These should all come pre-installed on GitHub runners now so we can save some CI time (~20s on Windows, ~13s on macOS).

I'm not sure if there's some other benefit to using the toolchain action, but we don't over in astral-sh/uv.

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -17,11 +17,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


indygreg (approved):
Installing the Rust toolchains in these workflows possibly predates the installation of Rust toolchains on GitHub Actions runners. There's no benefit to installing a custom toolchain: we just need a reasonably modern Rust install.

PR #319, opened 2024-09-06
Support for Python 3.13

zanieb:
A continuation of the wonderful work by @kpfleming in https://github.com/indygreg/python-build-standalone/pull/264

Adds initial support for Python 3.13 using CPython 3.13.0rc1.

There are a few caveats and interesting details:

- BOLT is disabled. There's a segmentation fault in the tests with the BOLT instrumented binaries. The BOLT optimizations are not critical, so we'll follow up on this separately. [See more context](https://github.com/zanieb/python-build-standalone/pull/6#issuecomment-2330386333).
- `mpdecimal` is now built from source on Windows. We already did this in Unix builds, but in Windows we were still using the bundled library. The bundled library is no longer used upstream and it seemed prudent to switch though it will remain available until 3.15. [See more context](https://github.com/zanieb/python-build-standalone/pull/6#issuecomment-2329887367).
- Apple cross-compilation is not available. I have a patch that adds support, but need to test it and it's not needed for these builds. [See more context](https://github.com/zanieb/python-build-standalone/pull/9/commits/447fb866d3198d619d359260d76886f3d42a18a1).
- `run_tests.py` was removed upstream. We provide a compatibility script that calls the appropriate command still so that our distributions are stable. We may want to change how `run_tests.py` is declared in the distribution metadata though. [See more context](https://github.com/indygreg/python-build-standalone/pull/319#discussion_r1747376558) .


cpython-unix/build-cpython.sh
```diff
@@ -134,9 +146,19 @@ if [ -n "${PYTHON_MEETS_MAXIMUM_VERSION_3_10}" ]; then
   patch -p1 -i ${ROOT}/patch-makesetup-deduplicate-objs.patch
 fi
 
+# testembed links against Tcl/Tk and libpython which already includes Tcl/Tk leading duplicate
+# symbols and warnings from objc (which then causes failures in `test_embed` during PGO).
+if [ -n "${PYTHON_MEETS_MAXIMUM_VERSION_3_13}" ]; then
+  patch -p1 -i ${ROOT}/patch-make-testembed-nolink-tcltk.patch
+fi
```
zanieb, line 153: See https://github.com/zanieb/python-build-standalone/pull/6#issuecomment-2329709954 for more context

cpython-unix/build-cpython.sh
```diff
@@ -426,6 +453,11 @@ if [ "${PYBUILD_PLATFORM}" = "macos" ]; then
     export MACOSX_DEPLOYMENT_TARGET="${APPLE_MIN_DEPLOYMENT_TARGET}"
 fi
 
+# ptsrname_r is only available in SDK 13.4+, but we target a lower version for compatibility.
+if [ "${PYBUILD_PLATFORM}" = "macos" ]; then
+    CONFIGURE_FLAGS="${CONFIGURE_FLAGS} ac_cv_func_ptsname_r=no"
+fi
```
zanieb, line 459: See https://github.com/zanieb/python-build-standalone/pull/6#issuecomment-2329715746 for more context.

cpython-unix/build-cpython.sh
```diff
@@ -742,7 +774,13 @@ s390x-unknown-linux-gnu)
...
     ;;
 x86_64-unknown-linux-*)
-    PYTHON_ARCH="x86_64-linux-gnu"
+    # In Python 3.13+, the musl target is identified in cross compiles and the output directory
+    # is named accordingly.
+    if [ "${CC}" = "musl-clang" ] && [ "${PYTHON_MAJMIN_VERSION}" = "3.13" ]; then
+        PYTHON_ARCH="x86_64-linux-musl"
+    else
+        PYTHON_ARCH="x86_64-linux-gnu"
+    fi
```
zanieb, line 783: See https://github.com/zanieb/python-build-standalone/pull/6#issuecomment-2330385678 for context.

zanieb:
It looks like I broke some older builds, which I stopped testing on in my fork to save time. I'll fix those quick.

cpython-unix/build-cpython.sh
henryiii: Python 3.13 only supports macOS 10.13+. The official build are built for 10.13+. So I'd think the macOS target version should be Python version dependent.

src/validation.rs
```diff
@@ -1470,14 +1506,25 @@ fn validate_extension_modules(
 
     if is_linux {
         wanted.extend(GLOBAL_EXTENSIONS_POSIX);
-        wanted.extend(GLOBAL_EXTENSIONS_LINUX);
+        if python_major_minor == "3.13" {
+            wanted.remove("_crypt");
```
charliermarsh, line 1510: Should these `wanted.remove` calls be encoded in constants too?

cpython-windows/build.py
```diff
@@ -1700,10 +1722,18 @@ def build_cpython(
             log("copying %s to %s" % (source, dest))
             shutil.copyfile(source, dest)
 
-        shutil.copyfile(
-            cpython_source_path / "Tools" / "scripts" / "run_tests.py",
-            out_dir / "python" / "build" / "run_tests.py",
-        )
+        # CPython 3.13 removed `run_tests.py`, we provide a compatibility script
+        # for now.
```
charliermarsh, line 1726: How is this used? Is it just something we invoke later on?

cpython-unix/build-cpython.sh
zanieb: I think we're intentionally setting a lower target deployment for compatibility with older operating systems. We do this in various other places already and CPython 3.13 has code to behave correctly if `ptsname_r` is not available.

We could consider bumping the minimum target version, but I think that should be separate from initial suport.

cpython-unix/extension-modules.yml
```diff
@@ -806,6 +856,7 @@ readline:
     - readline.c
   defines:
     - USE_LIBEDIT=1
+    - Py_RL_STARTUP_HOOK_TAKES_ARGS
```
charliermarsh, line 859: Can we document this?

zanieb: cc @kpfleming no clue what this is for

charliermarsh (approved):

cpython-windows/build.py
zanieb: We don't invoke it. It's placed in the distribution archive for consumers and there's a `run_tests` field in the `PYTHON.json` metadata which points to this script. This is briefly mentioned in the pull request summary, but I think we'll probably open an issue to just remove this in the future, allow it to be null in the metadata, or provide a command in the metadata instead of a path? All of these alternatives are technically breaking changes to the distributions though.

cpython-unix/extension-modules.yml
kpfleming: It's from here: https://github.com/python/cpython/blob/d343f977ba89c415aa4dcaf75122e0b4235d4fb1/configure.ac#L6517

kpfleming: And used here: https://github.com/python/cpython/blob/d343f977ba89c415aa4dcaf75122e0b4235d4fb1/Modules/readline.c#L1044

src/validation.rs
zanieb: Maybe, but notice we already split up the constants for LINUX but not POSIX and this is a much less invasive change. I'll look into addressing this separately, to avoid another CI run for something that's not blocking here.

cpython-unix/extension-modules.yml
zanieb: And we need to set it because... our editline version takes arguments?

src/validation.rs
zanieb: Adding a TODO for this since I need to push a commit anyway.

cpython-unix/extension-modules.yml
zanieb: I think I get this now. Added a comment.

zanieb:
I have a fix for the [Windows Python 3.8 failures](https://github.com/indygreg/python-build-standalone/actions/runs/10740463968/job/29788718089?pr=319) but will wait to push until I see if anything else fails.

cpython-windows/build.py
zanieb: For tracking, here was the first commit I did to get past this change https://github.com/indygreg/python-build-standalone/pull/264/commits/5afde6d9fd0272cb94724f26bf58f5ee2d42e6a4 — here we resolve the TODO there and provide the compatibility script.

zanieb: I tested the script on Linux, but not Windows.

cpython-unix/build-cpython.sh
henryiii: https://pythoninsider.blogspot.com/2024/06/python-3130-beta-2-released.html?m=1

> The minimum supported macOS version was changed from 10.9 to 10.13 (High Sierra). Older macOS versions will not be supported going forward.

zanieb: Well, that's good to know. Regardless, I'm not sure that means that we need to drop support yet. I actually tried to bump the minimum version first and it didn't resolve the commented issue because `ptsname_r` is only available in 10.13.4 not 10.13.0. Changing to a minimum supported patch version is a larger change and one I'm hesitant about it. Separately, there's not support for conditional macOS target versions yet and I'd need to implement that. I'd rather get this out the door first.

zanieb: Also the section you linked references the changes to "the macOS installer we provide" which is bit different than CPython itself no?

henryiii: Might be a question for @ned-deily. I believe officially only 10.13+ is tested and supported now, though it might happen to work for older versions for a while.

zanieb: Since we package the distributions here it should be fairly obvious when it stops working and we can address the problems then, either with patches or by upgrading to a newer version. I think the spirit here is to maintain compatibility with as many systems as feasible, so I'll bias towards that to start.

zanieb: cc @hugovk (if you're interested)

ned-deily: The comment about changing the minimum support macOS version from 10.9 to 10.13 does refer to the python.org macOS installer and not necessarily to CPython itself. (At the moment, we don't have an officially published support policy for which macOS releases are supported by CPython although we plan to make one.) But there is no intent to deliberately break support on older releases. The above patch, though, does point out a deficiency in CPython today as @zanieb notes, as there currently is no run-time test for the presence of `ptsname_r`, whose use was introduced fairly recently, which would be needed to maintain full weak-linking support on older systems. It would be good to open a cpython issue about that.

zanieb: Thanks for weighing in @ned-deily!

zanieb: Here's an upstream issue https://github.com/python/cpython/issues/123797

indygreg: I would be fine dropping 10.9 support on 3.13. I think targeting what the official installers support is _good enough_.

If some random user on 10.9-10.2 complains, they can always build things from source. Presumably they've been doing that for years because the ~entire world has dropped support for their OS version years ago.

indygreg (commented):
I skimmed this and it looks great!

I didn't thoroughly review the new variant CPython patches. Those can be tricky as the CPython build system is a moving target and we're doing some questionable things with some of these patches. But the Rust distribution validation has gotten pretty robust over the years and if things pass, I feel pretty confident about correctness.

FeodorFitsner:
[Python 3.13.0rc2 is out](https://blog.python.org/2024/09/python-3130rc2-3126-31110-31015-3920.html).

Is 3.13 pre-release binaries going to be published to releases or should be downloaded from build artifacts?

zanieb:
@FeodorFitsner we'll probably release them soon.

I'll work on rc2 support next week.

indygreg:
There was a joint release of all supported CPython versions a few minutes ago. Includes a ton of security fixes. We should try to get a release out soon.

https://www.python.org/downloads/release/python-3126/

zanieb:
👍 I'll work on all those next week then haha

zanieb:
Also of note

> Also, as mentioned in the previous release of 3.12, this release drops support for macOS versions 10.9 through 10.12. Versions of macOS older than 10.13 haven't been supported by Apple since 2019, and maintaining support for them has become too difficult. (All versions of Python 3.13 have already dropped support for them.)

indygreg:
Yes this project has had no problem building them. Actually testing on 10.9 and verifying builds work is another story, of course.

zanieb:
As a heads up to avoid duplicate work, I started bumping the patch versions https://github.com/indygreg/python-build-standalone/pull/322

ned-deily:
> Yes this project has had no problem building them. Actually testing on 10.9 and verifying builds work is another story, of course.

FTR, the release notice was a bit terse about that. We currently don't have any problem building Python for 10.9 and manually testing it. In fact, it's because of that we discovered at the last minute during the previous, 3.12.5, release, that the latest release of [pip 24.2 does not work on older macOS versions](https://github.com/pypa/pip/issues/12901) and that was a release blocker. Since it was at the last minute, we added a partial workaround to try to pin pip to its previous release on older systems but there have been and will be more issues like this so we decided to bite the bullet and bump to 10.13 for this current 3.12 release.

PR #323, opened 2024-09-07
Add patch for Apple cross compilation on 3.13+

zanieb:
Following up on #319 — adds a patch to support cross-compilation as we do for previous Python versions.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -73,8 +73,7 @@ pushd Python-${PYTHON_VERSION}
 
 # configure doesn't support cross-compiling on Apple. Teach it.
 if [ "${PYTHON_MAJMIN_VERSION}" = "3.13" ]; then
-    # TODO: Add support for cross-compiling on 3.13
-    :
+    patch -p1 -i ${ROOT}/patch-apple-cross-3.13.patch
 elif [ "${PYTHON_MAJMIN_VERSION}" = "3.12" ]; then
     patch -p1 -i ${ROOT}/patch-apple-cross-3.12.patch
 else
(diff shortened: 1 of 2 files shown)


zanieb:
@indygreg could you give me some context on testing this since it's not needed for CI to pass?

indygreg (approved):
You test this by passing `--target-triple` to `./build-macos.py` targeting the non-native target. So from Apple Silicon you want `--target-triple x86_64-apple-darwin` to cross-compile to x86-64. A working build is almost certainly sign of success.

To test from CI, you could switch the macOS runners to a new/old version to force cross-compiles. PGO and BOLT would stop working. But it otherwise should _just work_.

zanieb:
Ah okay that makes sense — I was wondering why CI didn't cross compile if we were patching for it and thought maybe this was for some other target.

This fails for me:

> cpython-3.13> configure: error: cross build not supported for x86_64-apple-darwin

I'll play with the patch some more.

indygreg:
CPython's build system has historically made some questionable decisions about how to build on/for Apple platforms. e.g. they've historically not supported cross-compiling (unless doing a multi-arch "universal" build) despite cross-compiling for Apple platforms being probably one of the easiest platforms to cross-compile for!

My guess is they tweaked more configure checks in this area (possibly as part of shoring up iOS support) and created more work for us in the process.

We don't need to support cross-compiles anymore. Historically I needed it because GitHub Actions didn't have macOS ARM runners. At the moment we have x86-64 and ARM macOS runners and all is well.

zanieb:
It actually is really nice to be able to cross-compile so I can test the x86_64 builds from my laptop. Our builds are also slowed down by GitHub's concurrent macOS job limit (only 5). There are alternative platforms that provide macOS runners that would let us have a higher concurrency limit, but they only support ARM. I'll probably try to fix it slowly.

zanieb:
Alright turned out to be pretty easy, works locally.

PR #328, opened 2024-09-09
Switch from `semver::Version` to `pep440_rs::Version` for Python version comparisons during release

zanieb:
Otherwise, we fail to parse `3.13.0rc2` because the Python prerelease syntax does not match the Cargo SemVer syntax

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,6 +1243,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "pep440_rs"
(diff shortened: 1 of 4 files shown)


zanieb:
This caused a dry-run of the release process to fail and is blocking the Python 3.13 release. I confirmed the release process works now locally.

charliermarsh (approved):

konstin (approved):

PR #336, opened 2024-10-02
Include free-threaded builds in the release artifacts

zanieb:
We need conditional build suffix support for the free-threaded builds.

Tested with:

```
just release-dry-run <token> f9203c323d8a4d10e3c64f29356c96f0e3907971 20241002
```

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -31,6 +31,46 @@ pub struct TripleRelease {
     pub install_only_suffix: &'static str,
     /// Minimum Python version this triple is released for.
     pub python_version_requirement: Option<VersionSpecifier>,
+    /// Additional build suffixes to release conditional on the Python version.
+    pub conditional_suffixes: Vec<ConditionalSuffixes>,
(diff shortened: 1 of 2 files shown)


src/release.rs
```diff
@@ -31,19 +31,68 @@ pub struct TripleRelease {
...
 pub static RELEASE_TRIPLES: Lazy<BTreeMap<&'static str, TripleRelease>> = Lazy::new(|| {
     let mut h = BTreeMap::new();
 
     // macOS.
     let macos_suffixes = vec!["debug", "pgo", "pgo+lto"];
+    let macos_suffixes_313 = vec![
+        "freethreaded+debug",
+        "freethreaded+pgo",
+        "freethreaded+pgo+lto",
+    ];
```
zanieb, line 85: Needs to be static so we can't derive it from the above list without a more invasive change.

charliermarsh (approved):

PR #340, opened 2024-10-07
Drop Python 3.8 artifacts from the release

zanieb:
We need more disk space for 3.13t and this seems like the most straight-forward way to get it?

See failure at https://github.com/indygreg/python-build-standalone/actions/runs/11224124068

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -33,7 +33,7 @@ async fn fetch_artifact(
     repo: &str,
     artifact: WorkflowListArtifact,
 ) -> Result<bytes::Bytes> {
-    println!("downloading {}", artifact.name);
+    println!("downloading artifact {}", artifact.name);
 
     let res = client
         .actions()
@@ -106,6 +106,8 @@ pub async fn command_fetch_release_distributions(args: &ArgMatches) -> Result<()
                 .to_string(),
         )
         .build()?;
+
+    let release_version_range = pep440_rs::VersionSpecifier::from_str(">=3.9")?;
 
     let workflows = client.workflows(org, repo);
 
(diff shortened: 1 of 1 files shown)


src/github.rs
```diff
@@ -271,7 +293,7 @@ pub async fn command_fetch_release_distributions(args: &ArgMatches) -> Result<()
             let dest_path = produce_install_only(path)?;
 
             println!(
-                "releasing {}",
+                "prepared {} for release",
```
zanieb, line 296: I found this misleading so I changed some of the messages

src/github.rs
```diff
@@ -208,49 +210,69 @@ pub async fn command_fetch_release_distributions(args: &ArgMatches) -> Result<()
...
+                    "{} not in release version range {}",
+                    name, release_version_range
+                );
+                continue;
+            }
+
             // Iterate over `RELEASE_TRIPLES` in reverse-order to ensure that if any triple is a
             // substring of another, the longest match is used.
-            if let Some((triple, release)) =
+            let Some((triple, release)) =
```
zanieb, line 231: Best viewed with [unified whitespace](https://github.com/indygreg/python-build-standalone/pull/340/files?diff=unified&w=1) — this section is dedented.

src/github.rs
```diff
@@ -208,49 +210,69 @@ pub async fn command_fetch_release_distributions(args: &ArgMatches) -> Result<()
...
+            };
+
+            let python_version = pep440_rs::Version::from_str(parts[1])?;
+            if !release_version_range.contains(&python_version) {
+                println!(
+                    "{} not in release version range {}",
+                    name, release_version_range
+                );
+                continue;
+            }
```
zanieb, line 227: This is the key change

src/github.rs
```diff
@@ -208,49 +210,69 @@ pub async fn command_fetch_release_distributions(args: &ArgMatches) -> Result<()
 
             let name = zf.name().to_string();
 
+            let parts = name.split('-').collect::<Vec<_>>();
+
+            if parts[0] != "cpython" {
+                println!("ignoring {} not a cpython artifact", name);
+                continue;
+            };
```
zanieb, line 218: We later filter these out https://github.com/indygreg/python-build-standalone/blob/68a3e155a00835c5b8c5c2256eff5beaf65ab447/src/github.rs#L365 so we can avoid downloading them here to save disk space

zanieb:
Merging to attempt to get the 3.13.0 final release out — we can explore alternative solutions in the future. 3.8 is EOL now, but we might want to ship build improvements for a bit still?

PR #355, opened 2024-10-14
Bump strawberryperl from 5.28.1.1 -> 5.38.2.2

zanieb:
This is a 64-bit distribution but the previous version was 32-bit


indygreg (approved):
Part of the Windows build. If CI passes, it works.

PR #357, opened 2024-10-14
Bump zlib from 1.2.13 -> 1.3.1

zanieb:


indygreg (approved):
We need to be careful about this one since a lot of components link against libz and any ABI differences could cause breakage. But I think 1.3 should be ABI compatible with 1.2 so it may just work.

PR #359, opened 2024-10-14
Bump strawberryperl from 5.28.1.1 -> 5.32.1.1

zanieb:
Latest is 5.38.2.2 but it drops 32 bit support which we might require?


zanieb:
Looks like #355 succeeded 👍

PR #360, opened 2024-10-14
Bump sqlite from 3.46.0 -> 3.46.1

zanieb:


indygreg (approved):
SQLite is generally very safe to keep up to date.

PR #364, opened 2024-10-14
Bump binutils from 2.42 -> 2.43

zanieb:


indygreg (approved):
binutils upgrades generally just work. So LGTM assuming CI passes.

PR #365, opened 2024-10-14
Bump autoconf from 2.71 -> 2.72

zanieb:


indygreg (approved):
Some tools can have esoteric dependencies on specific versions of autoconf. LGTM if it passes CI.

indygreg (approved):
This should be relatively low risk: either builds work or they don't.

zanieb:
I expect this to pass — the last CI run had one flaked job. Figured I'd test 3.14 first though.

PR #354, opened 2024-10-14
Bump patchelf from 0.13.1 -> 0.18.0

zanieb:

PR #366, opened 2024-10-14
Bump patchelf from 0.13.1 -> 0.17.2

zanieb:


zanieb:
Failing with

> A compiler with support for C++17 language features is required

See https://github.com/astral-sh/python-build-standalone/pull/354#issuecomment-2413003822 for suggested resolution

PR #326, opened 2024-09-09
Add support for freethreaded builds of Python 3.13

zanieb:
Closes https://github.com/indygreg/python-build-standalone/issues/320

There are a couple important notes here:

- We exceeded the GitHub matrix size limit in the Linux build so the 3.13 jobs were split out into a separate matrix.
- There is no support for freethreaded musl builds because freethreading requires mimalloc which requires stdatomic which is not available in the much older LLVM we are using there. I am working on upgrading those to LLVM 18 separately.

PR #368, opened 2024-10-15
Fix freethreaded include on Windows

zanieb:
In all the changes of #326, I lost track of the critical change to actually enable free-threading on Windows — this was the only distribution I didn't verify manually since I don't have a Windows machine. 

See https://github.com/python/cpython/issues/112984 for details on Windows changes upstream.

Adds a post-build runtime check for this flag when relevant in CI to avoid future regressions https://github.com/indygreg/python-build-standalone/pull/368/commits/d440dc3d3e6c2319d1b027cc24dc6c645ce9b59a


charliermarsh (approved):

zanieb:
@charliermarsh checked the builds on his Windows machine, thanks!

PR #370, opened 2024-10-16
Add sccache to `pythonbuild`

orf:
I was looking at adding this to the actual cpython build, but it looks a bit more complex. I'm not sure how to pass `CC` and `CXX` to make [as required by the docs](https://github.com/Mozilla-Actions/sccache-action?tab=readme-ov-file#cc-code), and `build-cpython.sh` seems to do some checks that could interfere with this:

https://github.com/indygreg/python-build-standalone/blob/78e57bd18b48c357904b47a6ae6999ebb255c578/cpython-unix/build-cpython.sh#L119

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -14,8 +14,14 @@ concurrency:
 jobs:
   pythonbuild:
     runs-on: 'macos-13'
+    env:
+      SCCACHE_GHA_ENABLED: "true"
+      RUSTC_WRAPPER: "sccache"
     steps:
       - uses: actions/checkout@v4
+
+      - name: Run sccache-cache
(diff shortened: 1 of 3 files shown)


indygreg (commented):
I used to have sccache support for this project. I removed it because it didn't yield a meaningful speed-up. The CI for this project is bottlenecked by the volume of CPython build jobs. And individual jobs are frequently I/O bottlenecked from extracting tar archives into containers. There's some CPU bottlenecks compiling. But sccache fails to speed it up materially. I didn't think the complexity was justified.

zanieb:
Going to close per that comment, but if you have a demonstrable speedup in a fork I'm happy to reconsider — the slow CI is tough.

PR #371, opened 2024-10-16
Add build attestations. Closes #343

### LINKED ISSUE #343: Add release atteststions
Hello!

First, thanks for this fantastic project - it’s a great help to the community and ecosystem at large.

I would like to suggest adding [artifact attestations ](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds) to the releases.

This would enable supply chain verification for these builds, and provide a layer of validation above just verifying the signature.

it’s pretty simple to add: just a single step, with no configuration or changes required in the binary itself.

what do you think?

orf:
This adds build attestations for our builds, following the guides here: https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries

Closes #343

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -201,6 +201,12 @@ jobs:
             runner: macos-13
             py: 'cpython-3.13'
             options: 'freethreaded+pgo+lto'
+
+    permissions:
+      id-token: write
+      contents: read
+      attestations: write
+
     needs:
       - pythonbuild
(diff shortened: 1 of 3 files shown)


zanieb:
Thank you! Should we only be attesting the artifacts on `main`?

PR #373, opened 2024-10-17
Rename `python3.13t.exe` to `python.exe` in Windows free-threaded distributions

zanieb:
Addressing https://github.com/astral-sh/uv/issues/8298

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -1759,6 +1759,15 @@ def build_cpython(
             log("copying %s to %s" % (source, dest))
             shutil.copyfile(source, dest)
 
+        # Rename to `python.exe` when an alternative executable is built, e.g., when
+        # free-threading is enabled the name is `python3.13t.exe`.
+        canonical_python_exe = install_dir / "python.exe"
+        if not canonical_python_exe.exists():
+            os.rename(
+                install_dir / python_exe,
+                canonical_python_exe,
+            )
+
         # CPython 3.13 removed `run_tests.py`, we provide a compatibility script
         # for now.
         if meets_python_minimum_version(python_version, "3.13"):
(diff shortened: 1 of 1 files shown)


indygreg:
Reading the linked issue, I guess we're going the rename route instead of an additional copy?

On Linux and macOS before freethreaded builds the canonical binary was pythonX.Y and we had symlinks from pythonX and python to it. I have a light preference towards shipping copies of the exe to maintain the same logical layout. But if that's not how the official Windows installers work, then maybe the single python.exe is best.

zanieb:
> Reading the linked issue, I guess we're going the rename route instead of an additional copy?

Yeah, I initially thought the copy was a better option but I trust Sam and Steve's opinions.

It seems simplest to maintain consistency with the existing Windows distributions first, i.e., by using `python.exe`. I like the idea of matching the Linux and macOS layouts, but parity with Linux and macOS is tricky, since symlinks are generally a privileged operation on Windows. My experience thus far in uv is that Windows Python executables we discover have the same name across versions. I think they avoided this with the free-threaded executable in the official installer to avoid breaking people's systems, but we don't really have that problem here.

cpython-windows/build.py
```diff
@@ -1759,6 +1759,15 @@ def build_cpython(
             log("copying %s to %s" % (source, dest))
             shutil.copyfile(source, dest)
 
+        # Rename to `python.exe` when an alternative executable is built, e.g., when
+        # free-threading is enabled the name is `python3.13t.exe`.
+        canonical_python_exe = install_dir / "python.exe"
```
charliermarsh, line 1764: Is `install_dir` the same as `out_dir / "python" / "install"`?

charliermarsh: I'm just comparing to `python_exe = out_dir / "python" / "install" / python_exe` below.

cpython-windows/build.py
```diff
@@ -1759,6 +1759,15 @@ def build_cpython(
...
+        if not canonical_python_exe.exists():
+            os.rename(
+                install_dir / python_exe,
```
charliermarsh, line 1767: Would it make sense to instead test whether `python_exe` is `"python.exe"`?

cpython-windows/build.py
zanieb: Yep

https://github.com/indygreg/python-build-standalone/blob/00ea84cd2a525c8d20566482633903878bae2996/cpython-windows/build.py#L1655

cpython-windows/build.py
zanieb: I considered both option and don't see a strong case for either one. I think this approach avoids coupling assumptions about the name and mildly prefer it for that reason.

charliermarsh (approved):

PR #383, opened 2024-10-29
Disable cron for forks

hugovk:
Only run the scheduled cron for upstream:

https://dev.to/hugovk/til-how-to-disable-cron-for-github-forks-2d0l

This is a big CI matrix, we don't need to run the cron on forks, but contributors might nevertheless want run the CI before opening a PR (and clogging up the upstream CI queue).

---

Background: when pushing stuff to other repos, I was wondering why it was taking a long time for jobs to get a runner, but only in the afternoon, and for two or three hours.

After checking it's not a GitHub outage, I figured out it must be a cron from another repo I'd forked. But not so easy to find when you have over [1,000 forks](https://github.com/hugovk?tab=repositories&q=&type=fork&language=&sort=)! So I wrote a script to call the `gh` CLI to find which repo currently had a running CI, and surprise, surprise, it was this big 'un! :D

(As it happens, I got to use the same script a few days later to find another running in the morning!)

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -13,6 +13,7 @@ concurrency:
 
 jobs:
   pythonbuild:
+    if: ${{ github.repository_owner == 'indygreg' || github.event_name != 'schedule' }}
     runs-on: 'macos-13'
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


zanieb:
Thanks! This makes sense to me. I feel like we could also just disable the scheduled jobs entirely as I don't really use them in any way.

hugovk:
Okay, here we go! https://github.com/indygreg/python-build-standalone/pull/384

Issue #376, opened 2024-10-21
Windows CI failing after latest GHA update


zanieb:
I presume this is caused by a GHA bump

https://github.com/actions/runner-images/releases/tag/win19%2F20241015.1

A successful build (on the old runner image)

```
2024-10-16T20:43:49.9341964Z cpython>        FindVCRedistDir:
2024-10-16T20:43:49.9343225Z cpython>          VC Redist Directory: C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\\Redist\MSVC\14.29.30133\x64\
2024-10-16T20:43:49.9344590Z cpython>          VC Redist Version: 14.29.30133
2024-10-16T20:43:49.9739128Z cpython>        FindVCRuntime:
2024-10-16T20:43:49.9740432Z cpython>          VC Runtime DLL(s):
2024-10-16T20:43:49.9742801Z cpython>          - C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Redist\MSVC\14.29.30133\x64\Microsoft.VC142.CRT\vcruntime140.dll
2024-10-16T20:43:49.9745498Z cpython>          - C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Redist\MSVC\14.29.30133\x64\Microsoft.VC142.CRT\vcruntime140_1.dll
```

and a failing build (on the new runner image)

```
2024-10-21T17:20:23.9829945Z cpython>        (FindVCRuntime target) ->
2024-10-21T17:20:23.9833336Z cpython>          C:\Users\RUNNER~1\AppData\Local\Temp\python-build-1xvutigs\Python-3.9.20\PCbuild\pyproject.props(225,5): warning : vcruntime*.dll not found under C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\\Redist\MSVC\14.29.30156\x86\. [C:\Users\RUNNER~1\AppData\Local\Temp\python-build-1xvutigs\Python-3.9.20\PCbuild\pythoncore.vcxproj]
```

zanieb:
Perhaps related https://github.com/actions/runner-images/issues/10819

zanieb:
There was also a change to the Cygwin setup files on Oct 20

Various problems also being reported at https://sourceware.org/pipermail/cygwin/2024-October/256601.html

indygreg:
[...]

PR #384, opened 2024-10-29
Disable cron schedule

hugovk:
Follow on from https://github.com/indygreg/python-build-standalone/pull/383#issuecomment-2444237428.

This will save a lot of CI time each day.


zanieb:
(Windows failures unrelated of course, ref #376)

PR #385, opened 2024-10-29
Drop support for EOL Python 3.8

hugovk:
Fixes https://github.com/indygreg/python-build-standalone/issues/342.

This will help reduce the huge build matrix, which will need to grow fairly soon when 3.14 is added.


cpython-windows/build.py
```diff
@@ -614,230 +614,6 @@ def hack_project_files(
         pass
 
 
-PYPORT_EXPORT_SEARCH_39 = b"""
```
hugovk, line 617: The use of these constants were removed in https://github.com/indygreg/python-build-standalone/issues/342.

zanieb:
Note I already removed this from the build matrix in #344

zanieb:
@hugovk would you update with `main` for the Windows fixes?

zanieb (approved):
Awesome

PR #391, opened 2024-11-11
Require mimalloc explicitly on all GNU Python 3.13 builds

zanieb:
Previously, it was only explicitly required for the freethreaded builds

Closes https://github.com/indygreg/python-build-standalone/issues/389

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -355,8 +355,14 @@ if [ -n "${CPYTHON_DEBUG}" ]; then
     CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --with-pydebug"
 fi
 
+# Explicitly enable mimalloc on 3.13+, it's already included by default but with this it'll fail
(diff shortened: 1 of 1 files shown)


colesbury:
The `--with-mimalloc` option is the default, so I think 3.13 it's already built with it. The configure option just adds *build time support*. You need to set the environment variable `PYTHONMALLOC=mimalloc` to use mimalloc in the default build. (The free threading build requires and always uses mimalloc.)

I ran `PYTHONMALLOC=mimalloc uv run -p 3.13 python -c "while True: x = object()"` locally and I see mimalloc functions being executed in `perf top`.

zanieb:
Thanks for the info! It's the default if it's available, which it should be for most of our builds but it might be nice to make it explicit so it fails if it's not available for some reason. Notably the musl builds don't have a sufficient LLVM version for mimalloc and fail here — this is a known issue, ref https://github.com/indygreg/python-build-standalone/pull/326#issuecomment-2375558046.

PR #393, opened 2024-11-21
Remove redundant version check

hugovk:
Follow on from https://github.com/indygreg/python-build-standalone/pull/385.

`matches!(python_major_minor, "3.9" | "3.10" | "3.11" | "3.12" | "3.13")` is essentially everything and tautologous, so let's remove it.
 
Also add colour to CI logs for readability.


zanieb (approved):
Thanks!

PR #400, opened 2024-12-04
Update to latest Python patch versions

### LINKED ISSUE #398: Python 3.13.1, 3.12.8, 3.11.11, 3.10.16 and 3.9.21
The Python versions in $title were released today. They contain security fixes. We should try to roll a release as soon as reasonable.

zanieb:
Closes #398


indygreg (approved):
Looks pretty straightforward (as it should be).

Please also check for pip, setuptools, openssl, and SQLite updates. Those are the most common during routine upgrades.

zanieb:
See

- #401 
- #402

No OpenSSL patches

zanieb:
cc @indygreg it looks like now the `mips` 3.13 builds pull `-latomic` from the system?

```
File "/home/runner/work/python-build-standalone/python-build-standalone/cpython-unix/build.py", line 549, in python_build_info
    raise Exception("unexpected library in LIBS (%s): %s" % (libs, lib))
Exception: unexpected library in LIBS (['-lpthread', '-ldl', '-lpthread', '-lutil', '-latomic']): atomic
```

I'm not sure what changed there, but I believe it is correct to allow it https://github.com/indygreg/python-build-standalone/pull/400/commits/59d8fab88aab666d703278ef0969bdc616b7a0f0

PR #406, opened 2024-12-06
Copy the Python 3.13t executable to the canonical name instead of renaming

zanieb:
As the simplest immediate fix for https://github.com/indygreg/python-build-standalone/issues/405

The main downside here is increased artifact size. I think that's okay for now, we can remove the extra binary in the future once we resolve the root cause. It's only 89K.


zanieb:
I'll test this artifact on a Windows machine — previously we did not test a virtual environment.

shauneccles:
I downloaded `cpython-3.13-vcvars32.bat-freethreaded+pgo.zip` from the GHA run

All appears working on Windows 11 FWIW

image

charliermarsh (approved):

PR #411, opened 2024-12-06
Scope `libatomic` allowance to MIPS on Python 3.13

zanieb:
Addressing the initial feedback in https://github.com/indygreg/python-build-standalone/issues/410

### DIFF
--- a/cpython-unix/build.py
+++ b/cpython-unix/build.py
@@ -59,7 +59,6 @@ LINUX_ALLOW_SYSTEM_LIBRARIES = {
     "pthread",
     "rt",
     "util",
-    "atomic",
 }
 MACOS_ALLOW_SYSTEM_LIBRARIES = {"dl", "m", "pthread"}
 MACOS_ALLOW_FRAMEWORKS = {"CoreFoundation"}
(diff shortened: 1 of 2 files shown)


indygreg (approved):
Thank you.

A proper solution here might be to statically link libatomic. But that can be done separately.

zanieb:
Thanks for giving it a look

> A proper solution here might be to statically link libatomic. But that can be done separately.

Agreed. I've started looking into that, but after engaging upstream (https://github.com/python/cpython/issues/127708#issuecomment-2524172362) I'm starting to wonder about the value of the MIPS distributions?

indygreg:
Let's merge this since it fixes the `PYTHON.json` metadata advertisement for `libatomic.so` on Linux (which is wrong). We can discuss MIPS elsewhere.

indygreg:
Also, `extension-modules.yml` has the reference to `libatomic` regarding OpenSSL I was thinking ok:

```
_hashlib:
  sources:
    - _hashopenssl.c
  includes-deps:
    - include
  links:
    - crypto
  links-conditional:
    # Clang built OpenSSL 3.x depends on libatomic. We force linking against
    # a static library so there isn't a runtime dependency.
    - name: ':libatomic.a'
      targets:
        - i686-unknown-linux-gnu
        - mips-unknown-linux-gnu
        - mipsel-unknown-linux-gnu
        - x86_64.*-unknown-linux-gnu
```

We _could_ change the `build-cpython.sh` script to statically link `libatomic.a`.

zanieb:
Yeah I was looking at that implementation to guide how to hide the symbol and force it to statically link 👍

PR #414, opened 2024-12-08
Strip versioned Xcode path from build flags

charliermarsh:
## Summary

We get reports of users failing to build extension modules on macOS due to the presence of a hardcoded Xcode path. Here's one example:

bafkreih5iht5nuyhkuzrs3npkio5smggpcrwmwirjr7cq7wiftyx6f77xe-1

I wasn't able to reproduce this locally... until I set `CFLAGS` in my environment (to anything). It turns out that CPython already strips `-isysroot` on macOS, but _not_ if `CFLAGS` is set. See: https://github.com/python/cpython/blob/a03efb533a58fd13fb0cc7f4a5c02c8406a407bd/Lib/_osx_support.py#L331.

(I find this a bit surprising -- it should probably _still_ strip the values that are provided via `_sysconfigdata_t_darwin_darwin.py`, and only _not_ strip values that are set explicitly via a `CFLAGS` environment variable? But alas.)

This PR modifies our macOS builds to always strip `-isysroot`. We're effectively already doing this for users that _don't_ have any `CFLAGS` set globally, and the current value is almost always going to be wrong.

## Test Plan

I confirmed that `CFLAGS="-Wunreachable-code" uv pip install cffi==1.17.1 --python 3.13t --no-cache --verbose` succeeds with a build from this branch, but fails on `main`.


cpython-unix/build.py
```diff
@@ -165,23 +165,7 @@ def add_target_env(env, build_platform, target_triple, build_env):
...
-            # will give us the headers that Xcode is configured to use.
-            res = subprocess.run(
-                ["xcrun", "--sdk", sdk_platform, "--show-sdk-path"],
-                check=True,
-                capture_output=True,
-                encoding="utf-8",
-            )
-
-            sdk_path = res.stdout.strip()
+        sdk_path = os.environ["APPLE_SDK_PATH"]
```
indygreg, line 168: This may break local builds outside of CI which don't have `APPLE_SDK_PATH` set.

cpython-unix/build.py
```diff
@@ -817,6 +801,11 @@ def build_cpython(
         if "lto" in parsed_build_options:
             env["CPYTHON_LTO"] = "1"
 
+        sdk_path = os.environ["APPLE_SDK_PATH"]
+        if not os.path.exists(sdk_path):
+            raise Exception("macOS SDK path %s does not exist" % sdk_path)
+        env["APPLE_SDK_PATH"] = sdk_path
```
indygreg, line 807: This check is already performed around line 170. Do we need to check this again?

cpython-unix/build.py
charliermarsh: Ah ok -- I was partly trying to understand whether this is _always_ set in CI or if I was missing something. Makes sense.

cpython-unix/build.py
charliermarsh: Yeah will DRY this up once it's working as expected.

indygreg:
Alternatively, consider removing the sysroot argument from compiler flags. This is probably inferred by the compiler. If it isn't in a default CPython build's cflags, just rip it out.

charliermarsh:
Yeah, let me check. I'm trying to understand why it _doesn't_ appear for me when building `cffi` (unlike in the repro [here](https://bsky.app/profile/hynek.me/post/3lchmbq3zac2y)).

It's not present in `sysconfig.get_config_var`:

```python
>>> sysconfig.get_config_var("CFLAGS")
'-fno-strict-overflow -Wsign-compare -Wunreachable-code -DNDEBUG -g -O3 -Wall -arch arm64 -mmacosx-version-min=11.0 -Wno-nullability-completeness -Wno-expansion-to-defined -Wno-undef-prefix   -fPIC    -Werror=unguarded-availability-new'
```

But it does appear in `_sysconfigdata_t_darwin_darwin.py`:

```
'CFLAGS': '-fno-strict-overflow -Wsign-compare -Wunreachable-code -DNDEBUG -g -O3 -Wall -arch arm64 -mmacosx-version-min=11.0 -Wno-nullability-completeness -Wno-expansion-to-defined -Wno-undef-prefix  -isysroot /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk -fPIC    -Werror=unguarded-availability-new',
```

charliermarsh:
Interesting, it looks like CPython itself strips these out in some cases?

https://github.com/python/cpython/blob/a03efb533a58fd13fb0cc7f4a5c02c8406a407bd/Lib/_osx_support.py#L438

charliermarsh:
Ahhh ok. It seems that CPython does not strip `--isysroot` (even from the `sysconfig`-provided `CFLAGS`) _if_ you have a non-empty `CFLAGS` set in your environment, so I can reproduce the failure with:

```
CFLAGS="-Wunreachable-code" uv pip install cffi==1.17.1 --python 3.13t --no-cache --verbose
```

That's interesting. Should we just drop it then from `CFLAGS` et al?

charliermarsh:
It seems more correct to me to just drop `isysroot` everywhere, in that case.

indygreg:
the sysroot arguments are a bit nonconventional. I use those to target a specific Xcode toolchain instead of relying on environment variables. I prefer the arguments over environment variables because they are visible in compiler flags. Env variables are often not logged nor easily preserved by build systems, like Python's sysconfig.

charliermarsh:
Does it seem right to you to strip it from `_sysconfigdata_t_darwin_darwin.py`, though? (That's my latest attempt, above.)

charliermarsh:
(That way, we retain the use of arguments over environment variables in `python-build-standalone` itself, but avoid shipping those unwanted paths to users.)

charliermarsh:
Okay this work on Python 3.13, but earlier versions have their `_sysconfigdata_t_darwin_darwin.py` formatted such that `-isysroot` and the path appear in separate strings within an implicit concatenation, which is super annoying... I'm tempted to re-format them prior to find-replace.

cpython-unix/build-cpython.sh
```diff
@@ -628,6 +629,41 @@ def replace_in_all(search, replace):
...
+    build_time_vars = locals_dict['build_time_vars']
+
+    with open(SYSCONFIGDATA, "wb") as fh:
+        fh.write(b'# system configuration generated and used by the sysconfig module\n')
+        fh.write(('build_time_vars = %s' % json.dumps(build_time_vars, indent=4)).encode("utf-8"))
+        fh.close()
+
+
+# Format sysconfig to ensure that string replacements take effect.
+format_sysconfigdata()
```
charliermarsh, line 656: If we don't reformat the sysconfig data file, then `-isysroot` and the path that follows it appear on separate lines, as separate strings within an implicit concatenation -- and replacing them becomes _much_ harder. It looks like this file is already formatted on 3.13, but not on prior versions, so this seems fine IMO.

cpython-unix/build-cpython.sh
```diff
@@ -628,6 +629,41 @@ def replace_in_all(search, replace):
...
+    much harder. This function reformats the file to avoid this issue.
+
+    See: https://github.com/python/cpython/blob/a03efb533a58fd13fb0cc7f4a5c02c8406a407bd/Mac/BuildScript/build-installer.py#L1360C1-L1385C15.
+    """
+    with open(SYSCONFIGDATA, "rb") as fh:
+        data = fh.read()
+
+    globals_dict = {}
+    locals_dict = {}
+    exec(data, globals_dict, locals_dict)
```
zanieb, line 646: In general, I'd use [`runpy`](https://docs.python.org/3/library/runpy.html#runpy.run_path) for something like this — but I don't remember enough about the nuances to justify a switch.

Can you not just.. import the file as a module?

cpython-unix/build.py
```diff
@@ -186,6 +186,8 @@ def add_target_env(env, build_platform, target_triple, build_env):
         if not os.path.exists(sdk_path):
             raise Exception("macOS SDK path %s does not exist" % sdk_path)
 
+        env["APPLE_SDK_PATH"] = sdk_path
```
zanieb, line 189: What is this for?

zanieb: To pass through to `xcode_path = os.getenv("APPLE_SDK_PATH")`?

charliermarsh: Yeah that's right.

cpython-unix/build-cpython.sh
charliermarsh: We can... I was just copying the CPython version for parity though.

zanieb: Oh I see, sorry — we can stick with this.

zanieb (approved):

charliermarsh:
(Also re-tested these manually against https://github.com/astral-sh/uv/issues/9744.)

cpython-unix/build-cpython.sh
```diff
@@ -628,6 +629,41 @@ def replace_in_all(search, replace):
...
+    build_time_vars = locals_dict['build_time_vars']
+
+    with open(SYSCONFIGDATA, "wb") as fh:
+        fh.write(b'# system configuration generated and used by the sysconfig module\n')
+        fh.write(('build_time_vars = %s' % json.dumps(build_time_vars, indent=4)).encode("utf-8"))
```
indygreg, line 651: Does this need a sort_keys=True to make the output deterministic? If there are any dicts, it does.

indygreg (commented):
LGTM

PR #419, opened 2024-12-14
Build system refactor to prepare for parallel Python version building

indygreg:
This series of commits (read the individual commit messages for more details) prepares the cpython-unix build system for building multiple Python X.Y versions in parallel. We don't yet actually implement the frontend support for parallel builds so the scope of the PR is more manageable and limited to (supposedly) backwards compatible behavior.

### DIFF
--- a/cpython-unix/Makefile
+++ b/cpython-unix/Makefile
@@ -46,7 +46,7 @@ endif
 
 # Always write out settings files.
 $(shell $(RUN_BUILD) placeholder_archive makefiles)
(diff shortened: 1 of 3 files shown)


zanieb:
Thanks! Will review on Monday.

zanieb (approved):
The Make hurts my brain a little, but I have no objections :) thanks for the nice discrete commits

indygreg:
Yeah, make is a "fun" language.

PR #420, opened 2024-12-15
downloads: upgrade LLVM 18.0.8 -> 19.1.0

indygreg:


indygreg:
This looks to be passing CI. So asking for review.

charliermarsh (approved):

charliermarsh:
It looks like two builds failed here -- possible that those are new failures?

charliermarsh:
(Only asking because they're also failing on main and on my own PRs now :))

indygreg:
Uh what. I merged on my phone and I only saw green. Feel free to revert I guess.

indygreg:
I reproduced the failure locally and reverted.

This looks like upstream issue https://github.com/python/cpython/issues/124948.

I was going to say it was strange it only reproduced on 3.12. But we currently only have BOLT enabled on 3.12. So looks like a general BOLT regression on LLVM 19.1.0.

indygreg:
Well, not the same symbol as the upstream CPython issue. But same failure mode of unable to locate an address.

My best guess is BOLT isn't able to follow a rewrite performed by PGO or itself.

PR #421, opened 2024-12-15
Patch `_tkinter.c` to look in base prefix

charliermarsh:
## Summary

This PR attempts to adresss one of the [known quirks](https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html#tcl-tk-support-files) around Tcl/tk files. On Windows, CPython will look for the Tcl/tk files at `{base_prefix}/lib`. That works well for python-build-standalone, since we ship the files in that location. However, on Unix, CPython does _not_ check that location -- instead, for python-build-standalone, it just looks relative to the executable. This works for non-virtual environments, since CPython _will_ look in `{prefix}/lib`. But not for virtual environments.

The patch applied here ports the Windows discovery logic to Unix, so that we always hook up the right paths.

The advantage of this approach (vis-a-vis others that were considered) is that it doesn't rely on injecting an environment variable into the user's environment.

For now, I've only patched Python 3.13. Unfortunately, I'll need a _slightly_ different patch for each Python minor. I've already authored them locally, but I want to hold off on wiring them all up until we're aligned on this general approach.

In theory, I think this change could be upstreamed.

## Test Plan

- Build Python locally.
- `./python/install/bin/python -c "from tkinter import Tk; Tk()"`
- `./python/install/bin/python -m venv .venv`
- `.venv/bin/python  -c "from tkinter import Tk; Tk()"`


charliermarsh:
(To clarify, I'm looking for sign-off on the approach and input on whether we want to apply a separate patch for each Python minor, or cutoff support at some later version, like 3.11, or 3.12, or 3.13.)

indygreg (commented):
Historically I've tried to ensure that all Python X.Y versions behave the same. So nominally we should carry patch variants for all versions, as necessary.

I agree that this is the type of patch that upstream may want to take. However, I could see aspects of it getting stalled up. e.g. I wouldn't be surprised if some distros didn't install things at `/usr/lib/tcl` and the hardcoding of this path is probably _wrong_. (This is the type of thing that should be controllable via a configure argument and the actual path should also be sniffed via configure.)

I glanced at the patch and looking for the tcl files in `<prefix>/lib/tcl` should be reasonable for PBS. So the new functionality passes the smell test.

cpython-unix/build-cpython.sh
```diff
@@ -179,6 +179,15 @@ else
     patch -p1 -i ${ROOT}/patch-ctypes-callproc-legacy.patch
 fi
 
+# On Windows, CPython looks for the Tcl/Tk libraries relative to the base prefix,
+# with we want. But on Unix, it doesn't. This patch applies similar behavior on Unix,
```
zanieb, line 183: "which we want"?

charliermarsh:
Sounds good. I'll go ahead and add patches for the remaining minors.

Yeah, I briefly chatted with @zooba about this in DMs, just to ask for any context on why the Tcl/Tk behavior differs, and he suggested that it could make sense for us to add a build flag in CPython to enable standalone behavior, since it tends to be easier to get patches merged if they're not enabled by default and therefore don't impact redistributors.

zanieb (approved):

cpython-unix/patch-tkinter-3.13.patch
```diff
@@ -0,0 +1,127 @@
...
++            char *path;
++
++            if (!getenv("TCL_LIBRARY")) {
++                str_path = _get_tcl_lib_path();
++                if (str_path == NULL && PyErr_Occurred()) {
++                    Py_DECREF(m);
++                    return NULL;
++                }
++                if (str_path != NULL) {
++                    setenv("TCL_LIBRARY", path, 1);
```
carljm, line 112: It looks like `path` is used un-initialized here?

(Same in other versions.)

cpython-unix/patch-tkinter-3.13.patch
```diff
@@ -0,0 +1,127 @@
...
++                    set_var = 1;
++                }
++            }
++
+             Tcl_FindExecutable(PyBytes_AS_STRING(cexe));
++
++            if (set_var) {
++                unsetenv("TCL_LIBRARY");
```
carljm, line 120: Should `path` be freed here? (If it were ever allocated/initialized above, which it doesn't seem to be.) The equivalent in the Windows code path (`wcs_path`) is freed here.

(Same in other versions.)

carljm (commented):
I don't know anything about Tcl/Tk, so I don't fully understand what this code needs to do, but commented on a couple things I noticed. Couple other things:

* It seems like the return value of `_get_tcl_lib_path` is never decrefed and is just leaked? But if this is only called in one-time init code paths, probably nobody cares, one-time leaks are normal. And it looks like that was already the case in the Windows-only codepaths.
* It seems like some code duplication between Windows and non-Windows could probably be reduced here, but for now this keeps the patches simpler. Might be worth doing that if this is upstreamed in future.

charliermarsh:
Thank you @carljm!

PR #426, opened 2024-12-17
Switch to Depot runners in CI

zanieb:
Otherwise, the jobs here will consume our concurrency limits and block CI organization-wide. The Depot runners will give us greater concurrency limits, and notably, avoid colliding with our GitHub Runner limits — which are very low for macOS in particular.

~Also, uses larger GitHub Actions runners for Windows.~ I've reverted this because there's a new, inexplicable [failure for 32-bit 3.12](https://github.com/astral-sh/python-build-standalone/actions/runs/12381299627/job/34559706602?pr=426) 

Note that Depot doesn't support x86_64 macOS so we're now cross-compiling those. I previously verified the artifact was correct.

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -46,80 +46,62 @@ jobs:
       matrix:
         build:
           - target_triple: 'aarch64-apple-darwin'
-            runner: macos-14
             py: 'cpython-3.9'
             options: 'debug'
           - target_triple: 'aarch64-apple-darwin'
-            runner: macos-14
             py: 'cpython-3.9'
             options: 'pgo'
           - target_triple: 'aarch64-apple-darwin'
-            runner: macos-14
             py: 'cpython-3.9'
             options: 'pgo+lto'
 
           - target_triple: 'aarch64-apple-darwin'
-            runner: macos-14
             py: 'cpython-3.10'
             options: 'debug'
           - target_triple: 'aarch64-apple-darwin'
-            runner: macos-14
             py: 'cpython-3.10'
             options: 'pgo'
           - target_triple: 'aarch64-apple-darwin'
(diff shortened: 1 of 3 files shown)


hugovk:
Exciting!

One downside is non-Astral folk won't be able to test the CI on their forks because they won't have access to the private/paid runners.

We have a similar situation for CPython, we use Cirrus M1 runners for upstream and fallback to GitHub's `macos-14` runners for forks. Needless to say the YAML is a bit complex!

zanieb:
Ah yeah that's a good point I hadn't fully considered. I'll open an issue to track that.

zanieb:
Yeah it's really fun to see _all_ the Linux jobs run concurrently

zanieb:
That Windows job failed twice? But I don't see what could be different, we're still using a Windows 2022 runner. It's just a larger GitHub provided runner. This didn't fail in previous testing.

zanieb:
Alright, third time's the charm — something seems to be legitimately wrong with that specific combination.

charliermarsh (approved):

indygreg:
If Depot doesn't have macOS x86-64 runners, how does PGO work? Are the aarch64 macOS machines running x86-64 in emulation mode and using that to train the PGO?

x86-64 macOS is becoming vanishingly rare. (I'm typing this from a 2018 Intel MBP, which is barely usable.) So the loss of _real_ PGO on this platform may not be a huge deal. My experience with the x86-64 emulation is it is often fast/approximate enough. It's probably _good enough_. 🤷‍♂️

zanieb:
That's.. a good question. They must be running in emulation mode and I didn't notice because it's faster. Interesting.

I guess we should benchmark and see if there's a regression? I don't have the hardware to do so myself though.

PR #437, opened 2024-12-18
Add `ci:dry-run` to skip performing builds but still test the matrix

zanieb:
In general, I think this will be a little niche but in the short-term it'll be really helpful for iterating on our build matrix.

### DIFF
--- a/.github/workflows/apple.yml
+++ b/.github/workflows/apple.yml
@@ -90,6 +90,7 @@ jobs:
           path: build
 
       - name: Build
+        if: ${{ ! matrix.dry-run }}
         run: |
           if [ "${{ matrix.target_triple }}" = "aarch64-apple-darwin" ]; then
             export APPLE_SDK_PATH=/Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk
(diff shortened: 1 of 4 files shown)


ci-matrix.py
```diff
@@ -38,15 +38,19 @@ def parse_labels(labels: Optional[str]) -> dict[str, set[str]]:
...
             result["directives"].add("skip")
             continue
 
         if not label or ":" not in label:
             continue
 
         category, value = label.split(":", 1)
+
+        if category == "ci":
+            category = "directives"
```
zanieb, line 52: I could just avoid this rename and use `ci` instead of `directives` throughout, not sure if it really matters.

charliermarsh (approved):

.github/workflows/apple.yml
```diff
@@ -103,18 +104,22 @@ jobs:
...
 
       - name: Upload Distributions
+        if: ${{ ! matrix.dry-run }}
         uses: actions/upload-artifact@v4
         with:
           name: cpython-${{ matrix.python }}-${{ matrix.target_triple }}-${{ matrix.build_options }}
           path: dist/*
 
-      - uses: actions/checkout@v4
+      - name: Checkout macOS SDKs for validation
```
zanieb, line 113: (otherwise it's really confusing why there's a checkout action running here)

PR #445, opened 2024-12-19
Some cosmetic changes to the workflow names

zanieb:


.github/workflows/linux.yml
```diff
@@ -1,4 +1,4 @@
-name: Linux Python build
+name: linux
```
zanieb, line 1: Shortening these so we can see more of the target in the CI summary

.github/workflows/check.yml
```diff
@@ -1,7 +1,8 @@
-name: Check
+name: check
 
 on:
   push:
+    branches: [main]
```
zanieb, line 5: This check ran twice per commit

.github/workflows/macos.yml
```diff
@@ -1,4 +1,4 @@
-name: MacOS Python build
+name: macos
```
zanieb, line 1: Since it's "darwin" in the triple I'm sort of on the fence here. I could go either way.

zanieb: Perhaps the idea with `apple.yml` is that we'd be doing iOS, etc. cross-compiles? We could revisit at that point though.

PR #446, opened 2024-12-19
Use a larger runner for the release process

zanieb:
This has been failing repeatedly due to disk space, the hope is that this resolves it

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
   release:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-large
     steps:
       - uses: actions/checkout@v4
         with:


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
   release:
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-large
```
jsirois, line 29: You're likely aware there is a bunch of space taken up by default frameworks on the runners, but to add punch to that, I'll note Pex uses this: https://github.com/pex-tool/pex/blob/ed020784733d1233944774e9f1052490a48a28d0/.github/workflows/ci.yml#L111-L123

And it nets ~14GB currently: https://github.com/pex-tool/pex/actions/runs/12403252596/job/34628545010#step:2:326

I'm not sure if that is enough savings to help your case.

zanieb: Thanks! That's good to know.

indygreg: I'm a fan of the jlumbroso/free-disk-space action.

Issue #449, opened 2024-12-19
python-build-standalone ships sqlite3 module missing Connection.serialize, deserialize


dhandy2013:
I was really excited about switching away from `poetry` and `pyenv` to `uv` for building and testing our project.

However, our project relies on [sqlite3.Connection.serialize](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.serialize) and [sqlite3.Connection.deserialize](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.deserialize). The docs for these methods say:

> Note This method is only available if the underlying SQLite library has the deserialize API.

My system Python (Python 3.12 on Ubuntu 24.04) and pyenv-supplied Python 3.11 (also on Ubuntu 24.04) have these methods. It seems like this is something that should normally be available.

I really hate having to switch back to pyenv because of this issue. Can you compile sqlite3 with the `Connection.serialize` and `Connection.deserialize` methods?

Steps to reproduce:

I tested with python-build-standalone [cpython-3.11.11+20241206-x86_64-unknown-linux-gnu-install_only.tar.gz](cpython-3.11.11+20241206-x86_64-unknown-linux-gnu-install_only.tar.gz) on Ubuntu 24.04.

Steps to reproduce with python-build-standalone Python 3.11.11:
```
Python 3.11.11 (main, Dec  6 2024, 20:02:44) [Clang 18.1.8 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlite3
>>> sqlite3.sqlite_version
'3.47.1'
>>> from sqlite3 import Connection
>>> Connection.serialize
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'sqlite3.Connection' has no attribute 'serialize'
```

Comparing with pyenv Python 3.11:
```
Python 3.11.9 (main, Aug  6 2024, 18:48:15) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlite3
>>> sqlite3.sqlite_version
'3.45.1'
>>> from sqlite3 import Connection
>>> Connection.serialize
<method 'serialize' of 'sqlite3.Connection' objects>
```

zanieb:
Similar to https://github.com/astral-sh/python-build-standalone/issues/309 / https://github.com/astral-sh/python-build-standalone/pull/375, happy to enable features like this.

Do you know what flag is needed to do so?

[...]

PR #455, opened 2024-12-29
Include `libm` in statically linked sqlite3 builds

zanieb:
Looking into #449


zanieb:
Interestingly... on `main` we can see the logs

```
sqlite> checking SQL math functions... enabled
sqlite> checking for library containing ceil...
sqlite> -lm
```

but with https://github.com/astral-sh/python-build-standalone/pull/455/commits/026c992024a37770d7acb3ef292503721977acd5, we get

```
sqlite> checking SQL math functions... enabled
sqlite> checking for library containing ceil...
sqlite> none required
```

and... serialize / deserialize is still not available — I need to take a look at the pkg-config, I think.

PR #456, opened 2024-12-29
Add missing parser files to Python 3.13 Linux and macOS builds

### LINKED ISSUE #378: Object files missing?
Linking libpython 3.13 from object files in `cpython-3.13.0+20241016-x86_64_v2-unknown-linux-gnu-pgo-full.tar.zst` fails because of undefined `_PyToken_Init`, `_PyTokenizer_FromFile`, `_PyTokenizer_Get`, etc symbols.

Upon closer inspection, it looks like object files related to [`Parser/lexer`](https://github.com/python/cpython/tree/main/Parser/lexer) and [`Parser/tokenizer`](https://github.com/python/cpython/tree/main/Parser/tokenizer) are missing from `python/build/Parser`?

ntamas:
This PR fixes the CPython 3.13 distributions on Linux and macOS builds to include the additional object files in subdirectories of the `Parser` directory. Will fix #378 when merged.

Windows builds are unaffected by the bug because the Windows build script uses a different mechanism to collect the object files.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -989,7 +989,7 @@ ${BUILD_PYTHON} ${ROOT}/fix_shebangs.py ${ROOT}/out/python/install
 
 # Also copy object files so they can be linked in a custom manner by
 # downstream consumers.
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks!

PR #462, opened 2024-12-31
Upgrade LLVM to 19.1.6

indygreg:
This is a redo of #420, which was merged prematurely.

With the BOLT changes from #463 merged, LLVM 19 _just works_.

As part of this we also modernize the BOLT apply settings to follow the recommendations at https://llvm.org/devmtg/2024-03/slides/practical-use-of-bolt.pdf.

This includes enabling support for loading hot code from a huge page at runtime. This should _just work_ and could result in perf wins via improved iTLB hit rate, etc.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -259,9 +259,12 @@ if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_11}" ]; then
     patch -p1 -i ${ROOT}/patch-pwd-remove-conditional.patch
 fi
 
-# Adjust BOLT flags to yield better behavior. See inline details in patch.
 if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_12}" ]; then
+    # Adjust BOLT flags to yield better behavior. See inline details in patch.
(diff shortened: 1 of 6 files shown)


indygreg:
This is a redo of #420, which was merged prematurely.

The error we're seeing is:

```
BOLT-ERROR: unable to get new address corresponding to input address 0x3dd8f4 in function _PyEval_EvalFrameDefault. Consider adding this function to --skip-funcs=...
```

Looking at the LLVM source code, we see this error was introduced by https://github.com/llvm/llvm-project/pull/89681, which first shipped in LLVM 19.1.0.

The added test case says BOLT can't handle computed gotos in position independent code (PIC) (we use PIC since we build with -fPIC). `_PyEval_EvalFrameDefault` uses computed gotos if they are available. So LLVM 19.1 just isn't capable of BOLT optimizing `_PyEval_EvalFrameDefault`.

We'll need to add `_PyEval_EvalFrameDefault` to the BOLT skip list.

indygreg:
I diagnosed the BOLT issues and split those out into #463. I'm optimistic this will pass CI now.

zanieb (approved):

PR #463, opened 2025-01-01
unix: BOLT fixes

indygreg:
As part of investigating failures with BOLT when upgrading to LLVM 19, I found and fixed a few issues with BOLT.

First, `test_embed` had been segfaulting on BOLT instrumented binaries. Why I'm not entirely sure. But the segfault only seems to occur in instrumentation mode. These tests are doing low-level things with the interpreter. So I suspect some kind of global mutable state issue or something.

I found the exact tests triggering the segfaults and added annotations to skip them.

The CPython build system treats the segfault as fatal on 3.13 but not 3.12. This means that on 3.12 we were only running a subset of tests and not collecting BOLT instrumentation nor applying optimizations for all tests after `test_embed`.

The removal of the segfault enables us to enable BOLT on 3.13+.

Second, LLVM 19.x has a hard error when handling PIC compiled functions containing computed gotos. It appears prior versions of LLVM could silently have buggy behavior in this scenario. We need to skip functions with computed gotos to allow LLVM 19.x to work with BOLT. It makes sense to apply this patch before LLVM 19.x upgrade to prevent bugs with computed gotos.

Third, I noticed BOLT was complaining about the lack of `-update-debug-sections` during instrumentation.

The 2nd and 3rd issues require common arguments to both BOLT instrumentation and application invocations. The patch fixing both introduces a new configure variable to hold common BOLT arguments. This patch is a good candidate for upstreaming.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -253,6 +253,11 @@ if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_11}" ]; then
     patch -p1 -i ${ROOT}/patch-pwd-remove-conditional.patch
 fi
 
+# Adjust BOLT flags to yield better behavior. See inline details in patch.
+if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_12}" ]; then
+    patch -p1 -i ${ROOT}/patch-configure-bolt-flags.patch
+fi
+
 # The optimization make targets are both phony and non-phony. This leads
 # to PGO targets getting reevaluated after a build when you use multiple
 # make invocations. e.g. `make install` like we do below. Fix that.
@@ -285,6 +290,19 @@ autoconf
 # build system says we are cross-compiling.
 if [ -n "${CROSS_COMPILING}" ]; then
   patch -p1 -i ${ROOT}/patch-force-cross-compile.patch
+fi
+
+# BOLT instrumented binaries segfault in some test_embed tests for unknown reasons.
+# On 3.12 (minimum BOLT version), the segfault causes the test harness to
+# abort and BOLT optimization uses the partial test results. On 3.13, the segfault
+# is a fatal error.
+if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_10}" ]; then
+    patch -p1 -i ${ROOT}/patch-test-embed-prevent-segfault.patch
+fi
+
+# Same as above but for an additional set of tests introduced in 3.14.
+if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_14}" ]; then
+    patch -p1 -i ${ROOT}/patch-test-embed-prevent-segfault-3.14.patch
 fi
 
 # Most bits look at CFLAGS. But setup.py only looks at CPPFLAGS.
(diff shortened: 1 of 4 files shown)


indygreg:
Bleh. More `test_embed` failures on 3.14 that need resolved...

cpython-unix/build-cpython.sh
```diff
@@ -387,12 +405,7 @@ fi
 
 if [ -n "${CPYTHON_OPTIMIZED}" ]; then
     CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --enable-optimizations"
-    if [[ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_13}" && -n "${BOLT_CAPABLE}" ]]; then
-        # Due to a SEGFAULT when running `test_embed` with BOLT instrumented binaries, we can't use
-        # BOLT on Python 3.13+.
-        # TODO: Find a fix for this or consider skipping these tests specifically
-        echo "BOLT is disabled on Python 3.13+"
```
zanieb, line 394: Yay! Thank you.

cpython-unix/build-cpython.sh
```diff
@@ -287,6 +292,19 @@ if [ -n "${CROSS_COMPILING}" ]; then
   patch -p1 -i ${ROOT}/patch-force-cross-compile.patch
 fi
 
+# BOLT instrumented binaries segfault in some test_embed tests for unknown reasons.
+# On 3.12 (minimum BOLT version), the segfault causes the test harness to
+# abort and BOLT optimization uses the partial test results. On 3.13, the segfault
+# is a fatal error.
+if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_10}" ]; then
```
zanieb, line 299: Should these also be gated by `-n "${BOLT_CAPABLE}"` so we don't disable tests unnecessarily?

zanieb (approved):

cpython-unix/build-cpython.sh
indygreg: We don't actually run the full suite of tests. So it doesn't matter today.

(It would be nice to actually run the full test suite against the built distribution but that's for another PR I suppose.)

zanieb: Won't these be reflected in the distributed Python too? Like, a consumer would have these tests disabled?

Agree it doesn't seem critical.

indygreg: Oh, right, I forgot we distributed the unit tests.

I'll file an issue to clean things up. Agree we could do something better here. And we may be paving over a legit bug somewhere by disabling tests that segfault.

zanieb:
Sort of tragically, I cannot reproduce the `test_embed` failures in CPython alone so I'm going to have to start using components here one by one until it occurs? Or attach a debugger to this build? I'm not sure. It's looking hard.

zanieb:
Also, I'm not seeing anything related to 

> Second, LLVM 19.x has a hard error when handling PIC compiled functions containing computed gotos. It appears prior versions of LLVM could silently have buggy behavior in this scenario. We need to skip functions with computed gotos to allow LLVM 19.x to work with BOLT.

Am I missing some configuration to encounter this behavior?

zanieb:
Ah okay `--enable-shared` is needed to reproduce both of these failures — and second fix is required before I can get nice `test_embed` segfaults :)

Issue #238, opened 2024-03-20
Docker image builds hang on `apt-get update`


zanieb:
I'm seeing extremely slow containerized builds on Linux that hang at 100% CPU during `apt-get update`. I see very little network and disk IO and memory consumption.

Specifically, it looks like this is occurring for the `image-gcc`, `image-xcb`, and `-image-build` targets.

I can't figure out how to get verbose output from `apt` here, any recommendations would be helpful I've been spinning my wheels on it for a while.

zanieb:
I've tried things like...


```diff
❯ gd main -- base.Dockerfile
diff --git a/cpython-unix/base.Dockerfile b/cpython-unix/base.Dockerfile
index 76811a5..0401a9a 100644
--- a/cpython-unix/base.Dockerfile
+++ b/cpython-unix/base.Dockerfile
@@ -22,7 +22,8 @@ WORKDIR '/build'
 RUN for s in debian_jessie debian_jessie-updates debian-security_jessie/updates; do \
       echo "deb [trusted=yes] http://snapshot.debian.org/archive/${s%_*}/20230322T152120Z/ ${s#*_} main"; \
     done > /etc/apt/sources.list && \
-    ( echo 'quiet "true";'; \
+    ( echo 'Debug::Acquire::http "true";'; \
+      echo 'Debug::pkgAcquire::Worker "true";'; \
       echo 'APT::Get::Assume-Yes "true";'; \
       echo 'APT::Install-Recommends "false";'; \
       echo 'Acquire::Check-Valid-Until "false";'; \
@@ -33,4 +34,4 @@ RUN ( echo 'amd64'; \
       echo 'i386'; \
     ) > /var/lib/dpkg/arch
 
-RUN apt-get update
+RUN apt-get update 2>&1 | tee /tmp/output.txt
```

to get my hands on output from `apt`, works great with `docker build` but not with the build spawned by `ensure_docker_image`.

PR #468, opened 2025-01-02
unix: call `ulimit -n` to avoid overheads

indygreg:
See inline comment in `base.Dockerfile` for why this hack is necessary.

When upgrading my machine recently, the `rlim_max` value increased to ~1B, making `apt-get` invocations take multiple minutes to complete. https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1332440 seems to describe this issue.

The cross building images have a sufficiently new apt that doesn't seem impacted.

### DIFF
--- a/cpython-unix/base.Dockerfile
+++ b/cpython-unix/base.Dockerfile
@@ -33,4 +33,12 @@ RUN ( echo 'amd64'; \
       echo 'i386'; \
     ) > /var/lib/dpkg/arch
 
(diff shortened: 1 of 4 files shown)


zanieb:
Oh, I wonder if this solves https://github.com/astral-sh/python-build-standalone/issues/238

zanieb (approved):

indygreg:
I suspect this was the root cause of #238.

PR #469, opened 2025-01-02
unix: refresh test_embed patches to conditionally skip on BOLT

### LINKED ISSUE #465: Better handle BOLT skipped tests
See discussion at https://github.com/astral-sh/python-build-standalone/pull/463#pullrequestreview-2526931833.

tl;dr we had to disable some tests to avoid segfaults when BOLTing. We ship those test disables as part of the distributions, which feels wrong.

indygreg:
Turns out there is a support function to check for BOLT optimization. So we can conditionally skip tests on BOLT.

Closes #465.


zanieb (approved):
Thanks!

PR #472, opened 2025-01-03
Upgrade Python build environment requirements

zanieb:
Using `uv` instead of `pip-compile` so I can cross-compile the Windows requirements, upgrades the Python dependency lock files.


zanieb:
This also unblocks some weird zstandard build failure with cffi and Python 3.13 on my machine 🤷‍♀️

PR #479, opened 2025-01-06
Fix musl Linux release targets

zanieb:
- **Revert "Update release targets for x86_64 musl (#477)"**
- **Drop 3.14 from expected musl Linux release targets**


zanieb:
I misread the failure in #477

PR #481, opened 2025-01-07
Add support for build attestations

### LINKED ISSUE #343: Add release atteststions

samypr100:
Follow up on feedback from #371 and the previous work from @orf which I cherry-picked.

This adds build attestations for the builds using [actions/attest-build-provenance](https://github.com/actions/attest-build-provenance).

Closes #343

### Test Plan

Did a trial CI run which results in attestations like below for `cpython-3.10-aarch64-unknown-linux-gnu-lto`

* https://github.com/samypr100/python-build-standalone/attestations/4246020
* https://search.sigstore.dev/?logIndex=160192732


Gankra (approved):
Nice! I guess v2 dropped the need for contents: read?

PR #482, opened 2025-01-07
Fix x86_64-apple-darwin comment in targets.yml

kilian-hu:
Hi all,

just fixing an outdated comment about the min macOS version for the `x86_64-apple-darwin` target.

Thanks for maintaining this project!


zanieb (approved):
Thank you!

PR #483, opened 2025-01-07
Generate CI runners for build matrix dynamically

zanieb:


ci-matrix.py
```diff
@@ -201,9 +235,21 @@ def main() -> None:
...
+    with open(CI_RUNNERS_YAML, "r") as f:
+        runners = yaml.safe_load(f)
+
+    # If only free runners are allowed, reduce to a subset
+    if args.free_runners:
+        runners = {
+            runner: runner_config
+            for runner, runner_config in runners.items()
+            if runner_config.get("free")
+        }
```
zanieb, line 247: We can hook this up for forks, #431

ci-matrix.py
```diff
@@ -117,22 +120,50 @@ def generate_matrix_entries(
...
 
     # Create base entry that will be used for all variants
     base_entry = {
         "arch": arch,
         "target_triple": target_triple,
         "platform": platform,
+        "runner": runner,
+        # If `run` is in the config, use that — otherwise, default to if the
+        # runner architecture matches the build architecture
+        "run": str(config.get("run", runners[runner]["arch"] == arch)).lower(),
```
zanieb, line 166: Once we do #431, we might not have access to the runners needed for `run`. We'll have to do something a bit more complicated for PGO targets in that case but I don't handle this here.

ci-runners.yaml
```diff
@@ -0,0 +1,37 @@
...
+depot-ubuntu-22.04:
+  arch: x86_64
+  platform: linux
+  free: false
+
+# TODO: Enable this runner to perform native builds for aarch64
+# depot-ubuntu-22.04-arm:
+#   arch: aarch64
+#   platform: linux
+#   free: false
```
zanieb, line 12: Next, I'm going to try to enable PGO on aarch64 linux with this runner

charliermarsh (approved):

PR #484, opened 2025-01-07
Enable PGO and native builds for aarch64-unknown-linux-gnu

zanieb:


indygreg:
We currently don't use a modern LLVM toolchain for Linux aarch64. As a result we won't get the same optimizations from PGO or LTO.

It may just work though since the CPython configure should resolve the appropriate compiler flags from the detected toolchain.

We can look into adopting modern LLVM on Linux ARM separately.

zanieb:
@indygreg Are you sure? Perhaps I'm misunderstanding but it looks like we cross-compile with LLVM 19

https://github.com/astral-sh/python-build-standalone/blob/0be0a575b4a397599329aa11c919264361550d3b/pythonbuild/utils.py#L423-L436

https://github.com/astral-sh/python-build-standalone/actions/runs/12677576491/job/35333290679

zanieb:
Okay after some poking around this is blocked by the Dockerfile which cannot be built on `aarch64` yet.

zanieb:
See #670

Issue #388, opened 2024-10-30
Add RISC-V builds


lengau:
Related: https://github.com/indygreg/python-build-standalone/issues/219

We'd like to get RISC-V builds of Python. The current Docker images are intentionally built for an old version of Debian, so a new Docker image would need to be made for building Python on riscv64. Debian 10 (stretch) is the [first version that has the necessary gcc package](https://packages.debian.org/buster/gcc-riscv64-linux-gnu).

I believe the necessary steps would be:

1. Create a [riscv64 Dockerfile](https://github.com/indygreg/python-build-standalone/blob/main/cpython-unix/build.cross.Dockerfile) with Debian 10
2. Add the file to the [image builder](https://github.com/indygreg/python-build-standalone/blob/e8b031bf1a4852f1f3de1b6e51534749c6194e4c/.github/workflows/linux.yml#L51)
3. Add a risc-v option to [build-main.py](https://github.com/indygreg/python-build-standalone/blob/e8b031bf1a4852f1f3de1b6e51534749c6194e4c/cpython-unix/build-main.py#L104)
4. Add risc-v variants to [the build job](https://github.com/indygreg/python-build-standalone/blob/e8b031bf1a4852f1f3de1b6e51534749c6194e4c/.github/workflows/linux.yml#L873)

I can probably do this (and test it) at some point :-)

PR #492, opened 2025-01-10
Implement matrix sharding

zanieb:
Since we've encountered the GitHub matrix size limit of 256 entries a couple times now (recently in #388), this adds sharding to our matrix generation script.

There's now a `--max-shards` option which splits the matrix into (up to) the requested number of shards. We don't split it evenly, we eagerly fill each shard first. To consume the shards, we just copy the `build` job into two identical child jobs which consume each matrix shard. Ideally, we'd avoid code duplication here by using a re-usable action or something for the build-steps — but that's overkill at this point.

We only need this for Linux for the foreseeable future.


zanieb:
e.g., new warning and error

```
❯ uv run -q ci-matrix.py > matrix.json
warning: matrix of size 288 exceeds limit of 256 but sharding is not enabled; consider setting `--max-shards`
❯ uv run -q ci-matrix.py --max-shards=1 > matrix.json
error: matrix of size 288 requires 2 shards, but the maximum is 1; consider increasing `--max-shards`
```

zanieb:
e.g., we create empty shards when empty

```
❯ uv run -q ci-matrix.py --labels arch:foo --platform darwin --max-shards=2 | jq
{
  "0": {
    "include": []
  },
  "1": {
    "include": []
  }
}
```

PR #494, opened 2025-01-14
Add Python 3.14 support on Windows

zanieb:
The upstream build system now requires the latest tcl/tk and we want to upgrade anyway, so that happens here as well. There are some concerns about the new zlib/msvcrt dependencies, so I'm limiting these to the 3.14a to see how it goes.

See #495

### DIFF
--- a/ci-targets.yaml
+++ b/ci-targets.yaml
@@ -320,6 +320,7 @@ windows:
       - "3.11"
       - "3.12"
       - "3.13"
+      - "3.14"
     build_options:
       - pgo
     build_options_conditional:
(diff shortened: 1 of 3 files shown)


zanieb:
Note to self need to revert https://github.com/astral-sh/python-build-standalone/pull/478

charliermarsh (approved):

Gankra (commented):
Matches what we discussed, I maybe would have noted the msvcrt linkage is an artifact of the binary being built for mingw / windows-gnu, which is probably being done to cross-compile the binary from linux.

Issue #489, opened 2025-01-09
Spurious integrity mismatch for downloads


zanieb:
https://github.com/astral-sh/python-build-standalone/actions/runs/12637338024/job/35211538491

```
integrity mismatch on https://github.com/indygreg/toolchain-tools/releases/download/toolchain-bootstrap%2F20241231/llvm-19.1.6+20241231-gnu_only-x86_64-unknown-linux-gnu.tar.zst: wanted size=251277644, sha256=900bbb464d430e960ea97150e073395ae865332ce403e7c1e8f76640cab5e852; got size=127034607, sha256=9b5aac008b6c70df121b9e1958508082766a533225ae15ac7d8cdd776e949d14
make: *** [Makefile:84: /home/runner/work/python-build-standalone/python-build-standalone/build/llvm-19-x86_64-linux-19.1.6+20241231-linux64.tar] Error 1
```

https://github.com/astral-sh/python-build-standalone/actions/runs/12637338024/job/35211498463

```
pythonbuild.utils.IntegrityError: integrity mismatch on https://github.com/indygreg/toolchain-tools/releases/download/toolchain-bootstrap%2F20241231/llvm-19.1.6+20241231-gnu_only-x86_64-unknown-linux-gnu.tar.zst: wanted size=251277644, sha256=900bbb464d430e960ea97150e073395ae865332ce403e7c1e8f76640cab5e852; got size=127034607, sha256=9b5aac008b6c70df121b9e1958508082766a533225ae15ac7d8cdd776e949d14
make: *** [Makefile:84: /home/runner/work/python-build-standalone/python-build-standalone/build/llvm-19-x86_64-linux-19.1.6+20241231-linux64.tar] Error 1
```

zanieb:
Closing unless this occurs again.

zanieb:
Weird, failed again in https://github.com/astral-sh/python-build-standalone/actions/runs/12755334275/job/35551356915?pr=313

```
[...]

PR #496, opened 2025-01-14
Retry downloads when <75% of the file is downloaded

zanieb:
Closes https://github.com/astral-sh/python-build-standalone/issues/489 (hopefully)


Gankra (approved):
Hope it helps, at worst seems pretty harmless to add.

PR #501, opened 2025-01-19
Build attestations improvements

samypr100:
Per https://github.com/astral-sh/python-build-standalone/issues/343#issuecomment-2598666478

* Adds attestations to build-0 of the linux matrix (due to recent refactor)
* Adds attestations to release artifacts which include install only derived builds.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -245,6 +245,12 @@ jobs:
 
           build/pythonbuild validate-distribution ${EXTRA_ARGS} dist/*.tar.zst
 
+      - name: Generate attestations
+        uses: actions/attest-build-provenance@v2
+        if: ${{ github.ref == 'refs/heads/main' }}
+        with:
+          subject-path: dist/*
+
       - name: Upload Distribution
         if: ${{ ! matrix.dry-run }}
         uses: actions/upload-artifact@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,9 @@ env:
 permissions:
   contents: write
   packages: write
+  # Permissions used for actions/attest-build-provenance
+  id-token: write
+  attestations: write
 
 jobs:
   release:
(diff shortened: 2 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -245,6 +245,12 @@ jobs:
 
           build/pythonbuild validate-distribution ${EXTRA_ARGS} dist/*.tar.zst
 
+      - name: Generate attestations
```
zanieb, line 248: Whoops! Thank you!

.github/workflows/release.yml
```diff
@@ -75,3 +78,9 @@ jobs:
       # Uploading the relevant artifact to the GitHub release.
       - run: just release-run ${{ secrets.GITHUB_TOKEN }} ${{ github.event.inputs.sha }} ${{ github.event.inputs.tag }}
         if: ${{ github.event.inputs.dry-run == 'false' }}
+
+      - name: Generate attestations
+        uses: actions/attest-build-provenance@v2
+        if: ${{ github.event.inputs.dry-run == 'false' }}
+        with:
+          subject-path: dist/*
```
zanieb, line 86: I'm worried about the 1024 subject limit https://github.com/actions/attest-build-provenance?tab=readme-ov-file#subject-limits

zanieb: (1131 artifacts in the last release)

samypr100: Good catch. We could arguably divide the attestation groups by python versions.

samypr100: Adjusted the glob to skip .sha256 files as I don't think they benefit from attestation. Given this purely from a count perspective we can stay within the quota ~564

zanieb (approved):

geofft:
Something here seemed to not work quite right in [today's release](https://github.com/astral-sh/python-build-standalone/actions/runs/13165689836/job/36745165697):

```
Error: Error: Could not find subject at path dist/*.tar.@(zst|gz)
```

Is the extglob syntax not supported after all? I think the files exist....

samypr100:
> Something here seemed to not work quite right in [today's release](https://github.com/astral-sh/python-build-standalone/actions/runs/13165689836/job/36745165697):
> 
> ```
> Error: Error: Could not find subject at path dist/*.tar.@(zst|gz)
> ```
> 
> Is the extglob syntax not supported after all? I think the files exist....

Weird! I did try minimatch https://github.com/actions/toolkit/blob/main/packages/glob/src/glob.ts#L25 used by https://github.com/actions/attest/blob/main/src/subject.ts#L94 to verify the glob matched if the right files were in dist folder

PR #401, opened 2024-12-04
Upgrade pip and setuptools

zanieb:
- **Upgrade setuptools from 75.1.0 -> 75.6.0**
- **Upgrade pip from 24.1.2 -> 24.3.1**

PR #506, opened 2025-01-28
build-cpython.sh: Don't repack the pip wheel

geofft:
We're no longer patching it as of #401 (3dbca1c08f) so there's no longer a need to unpack and repack it.


zanieb (approved):
Nice, thank you!

geofft:
Ah, whoops, how do I do that? I think I may I need access to set labels.

PR #507, opened 2025-01-28
Make the installed pkg-config files relocatable (astral-sh/uv#11028)

geofft:
This ensures that e.g. `PKG_CONFIG_PATH=/wherever/python/lib/pkgconfig pkg-config --cflags` gets you existent paths in /wherever/python instead of nonexistent paths in /install.


lgarrison:
That was fast, thanks! I think `sed -i` -> `sed ${sed_args}` will unstick the MacOS CI.

geofft:
Ah, thanks, did not notice the script already did that :)

zanieb (approved):

PR #526, opened 2025-02-12
Patch 3.14.0a5 PGO on Windows amd64

zanieb:
See https://github.com/python/cpython/issues/130004

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -1342,7 +1342,7 @@ def build_cpython(
                         SUPPORT / "patch-disable-pgo-ceval-3.14.patch",
                     ],
                     str(cpython_source_path),
-                    env,
+                    env=os.environ,
                     exit_on_error=False,
                 )
 
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -1327,6 +1327,20 @@ def build_cpython(
         )
 
         if pgo:
+            if (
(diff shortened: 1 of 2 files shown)


cpython-windows/build.py
```diff
@@ -1327,6 +1327,25 @@ def build_cpython(
...
+            if (
+                meets_python_minimum_version(python_version, "3.14")
+                and build_platform == "x64"
+            ):
+                # On 3.14.0a5, PGO cannot be run on ceval so we disable it there
+                # See https://github.com/python/cpython/issues/130004
+                # We should be able to drop this patch in 3.14.0a6
+                exec_and_log(
+                    [
+                        "patch",
```
zanieb, line 1339: Presuming `patch` isn't available here and is why we don't have Windows patchfiles, but we'll see. Will follow with another approach if so.

Fidget-Spinner: `git apply` might work too if you have git.

Fidget-Spinner:
Seems like there's no need to apply the patch anymore. Hugo re-uploaded the 3.14a5 FTP tarballs with the MSVC patch applied.

hugovk:
Yep, the release needed three goes due to this Windows PGO bug and I accidentally uploaded the tarballs from the second attempt. Sorry about that!

zanieb:
Thank goodness, saves me even more time <3

PR #537, opened 2025-02-24
create install only archives for conditional versions

zsol:


zsol:
These are significantly smaller than the assets with all the build artifacts in them, so they're faster to download and eat less space. Over at https://github.com/zsol/dotslash-python I package some of these up into single file executables

zsol:
Thank you for shepherding these into `main`!

PR #541, opened 2025-02-26
Dynamically link libc in musl distributions

zanieb:


cpython-unix/build.py
```diff
@@ -121,7 +121,7 @@ def add_target_env(env, build_platform, target_triple, build_env):
             .replace("x86_64_v3-", "x86_64-")
             .replace("x86_64_v4-", "x86_64-")
             # TODO should the musl target be normalized?
-            .replace("-unknown-linux-musl", "-unknown-linux-gnu")
+            # .replace("-unknown-linux-musl", "-unknown-linux-gnu")
```
zanieb, line 124: The answer to this todo seems to be: no it should not generally be normalized like this because it causes problems. But then there are a couple specific builds where it _does_  need to be normalized because there isn't support for `-musl` targets?

zanieb:
I got the `x86_64-unknown-linux-musl-noopt` build passing on my machine and gave it a quick test in an alpine container and it looks like I didn't package it right.

```
/test # ldd ./python/install/bin/python
	/lib/ld-musl-x86_64.so.1 (0x7ebb73af0000)
Error loading shared library $ORIGIN/../lib/libpython3.13.so.1.0: No such file or directory (needed by ./python/install/bin/python)
	libc.so => /lib/ld-musl-x86_64.so.1 (0x7ebb73af0000)
Error relocating ./python/install/bin/python: Py_BytesMain: symbol not found
```

I presumed we just didn't package the .so, but it is there

```
/ # ls /test/python/install/lib/
Tix8.4.3              libpython3.13.so      pkgconfig             tcl8.6
itcl4.2.4             libpython3.13.so.1.0  python3.13            thread2.8.9
libpython3.13.a       libpython3.so         tcl8                  tk8.6
```

So I futzed with patchelf

```
/ # patchelf --set-rpath '/test/python/install/lib' /test/python/install/bin/python
/ # ldd /test/python/install/bin/python
	/lib/ld-musl-x86_64.so.1 (0x79dc36f38000)
Error loading shared library $ORIGIN/../lib/libpython3.13.so.1.0: No such file or directory (needed by /test/python/install/bin/python)
	libc.so => /lib/ld-musl-x86_64.so.1 (0x79dc36f38000)
Error relocating /test/python/install/bin/python: Py_BytesMain: symbol not found
```

Huh

```
/ # readelf -d /test/python/install/bin/python

Dynamic section at offset 0x388 contains 21 entries:
  Tag        Type                         Name/Value
 0x000000000000001d (RUNPATH)            Library runpath: [/test/python/install/lib]
 0x0000000000000001 (NEEDED)             Shared library: [$ORIGIN/../lib/libpython3.13.so.1.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so]
 0x000000000000000c (INIT)               0x401000
 0x000000000000000d (FINI)               0x401148
 0x0000000000000019 (INIT_ARRAY)         0x403e48
 0x000000000000001b (INIT_ARRAYSZ)       8 (bytes)
 0x000000000000001a (FINI_ARRAY)         0x403e50
 0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
 0x0000000000000004 (HASH)               0x3ff468
 0x000000006ffffef5 (GNU_HASH)           0x3ff428
 0x0000000000000005 (STRTAB)             0x3fd2e0
 0x0000000000000006 (SYMTAB)             0x3ff338
 0x000000000000000a (STRSZ)              167 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000015 (DEBUG)              0x0
 0x0000000000000003 (PLTGOT)             0x403fe8
 0x0000000000000002 (PLTRELSZ)           48 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x400470
 0x0000000000000000 (NULL)               0x0
```

I guess I'll try harder?

```
/ # patchelf --replace-needed '$ORIGIN/../lib/libpython3.13.so.1.0' '/test/python/install/lib/libpython3.13.so.1.0' /test/pyt
hon/install/bin/python
/ # /test/python/install/bin/python --version
Python 3.13.2
```

Seems like `$ORIGIN` probably isn't supported? Need to look into that.

zanieb:
Hm okay something is wrong with Python 3.11 — peachy otherwise!

zanieb:
This is loosely ready for review. I have some minor clean-ups and documentation to do.

indygreg (commented):
I strongly prefer retaining the ability to produce a statically linked musl distribution. There are some users of PyOxidizer who really like single file binaries. And you only get that if you statically link musl.

I would prefer standing up a new target variant with dynamic linked musl which is ABI compatible with the musl Python platform/target.

zanieb:
I can look into that. Will require #542.

cpython-unix/build-libX11.sh
```diff
@@ -75,6 +75,30 @@ if [ -n "${CROSS_COMPILING}" ]; then
...
+      ;;
+    i686-unknown-linux-musl)
+      EXTRA_FLAGS="${EXTRA_FLAGS} --enable-malloc0returnsnull"
+      ;;
+    mips-unknown-linux-musl)
+      EXTRA_FLAGS="${EXTRA_FLAGS} --enable-malloc0returnsnull"
+      ;;
+    mipsel-unknown-linux-musl)
+      EXTRA_FLAGS="${EXTRA_FLAGS} --enable-malloc0returnsnull"
+      ;;
```
zanieb, line 92: Probably a little overzealous including all these here, but of no real consequence.

pythonbuild/downloads.py
```diff
@@ -220,10 +220,10 @@
...
     },
     "musl": {
-        "url": "https://musl.libc.org/releases/musl-1.2.5.tar.gz",
-        "size": 1080786,
-        "sha256": "a9a118bbe84d8764da0ea0d28b3ab3fae8477fc7e4085d90102b8596fc7c75e4",
-        "version": "1.2.5",
+        "url": "https://musl.libc.org/releases/musl-1.2.2.tar.gz",
+        "size": 1055220,
+        "sha256": "9b969322012d796dc23dda27a35866034fa67d8fb67e0e2c45c913c3d43219dd",
+        "version": "1.2.2",
```
zanieb, line 226: We should probably retain the newer musl version for the static builds, as annoying as splitting it is.

src/validation.rs
```diff
@@ -1720,7 +1727,8 @@ fn validate_distribution(
 
     let is_debug = dist_filename.contains("-debug-");
 
-    let is_static = triple.contains("unknown-linux-musl");
+    // For now, there are now static builds — this is historic
+    let is_static = false;
```
zanieb, line 1731: Needs to change before merge

src/verify_distribution.py
```diff
@@ -53,7 +53,8 @@ def test_ctypes(self):
         import ctypes
 
         # pythonapi will be None on statically linked binaries.
-        if os.environ["TARGET_TRIPLE"].endswith("-unknown-linux-musl"):
+        is_static = False  # TODO: Populate if we have statically linked binaries again
```
zanieb, line 56: Needs to change before merge

src/validation.rs
```diff
@@ -71,6 +71,8 @@ const ELF_ALLOWED_LIBRARIES: &[&str] = &[
     "libpthread.so.0",
     "librt.so.1",
     "libutil.so.1",
+    // musl libc
+    "libc.so",
```
zanieb, line 75: Probably need to make this conditional on static

zanieb:
I seem to have broken the static builds. I'll continue to iterate on that.

cpython-unix/build.py
```diff
@@ -120,8 +120,6 @@ def add_target_env(env, build_platform, target_triple, build_env):
             target_triple.replace("x86_64_v2-", "x86_64-")
             .replace("x86_64_v3-", "x86_64-")
             .replace("x86_64_v4-", "x86_64-")
-            # TODO should the musl target be normalized?
-            .replace("-unknown-linux-musl", "-unknown-linux-gnu")
```
zanieb, line 124: See `xproto` and `xextproto` build changes

zanieb: The answer to this todo seems to be: no it should not generally be normalized like this because it causes problems. But then there are a couple specific builds where it does need to be normalized because there isn't support for `-musl` targets.

See also, the `tcl` build change.

pythonbuild/downloads.py
```diff
@@ -219,7 +219,15 @@
         "licenses": ["BSD-2-Clause"],
         "license_file": "LICENSE.mpdecimal.txt",
     },
+    # The musl toolchain for shared / dynamically linked builds
     "musl": {
+        "url": "https://musl.libc.org/releases/musl-1.2.2.tar.gz",
+        "size": 1055220,
+        "sha256": "9b969322012d796dc23dda27a35866034fa67d8fb67e0e2c45c913c3d43219dd",
+        "version": "1.2.2",
```
zanieb, line 227: We use an older version for dynamic builds, for compatibility with more systems. For example, the GitHub runners we're using did not have 1.2.5.

We retain the latest version for static builds, since compatibility isn't a concern.

cpython-unix/build-libX11.sh
geofft: I'm assuming we were previously implicitly doing this for musl?

geofft (approved):
LGTM. At first glance I'm okay changing the default musl build to be dynamic and to make the static variant an explicit option, and I would assume that works fine with the release process since it at most adds more artifacts, some of which might get missed, but the existing set of artifacts will still be present (with a semantics change for the musl ones). But we should probably double-check whether uv and Rye are chill with the change before shipping. Are there any other high-profile users we should check, especially ones we know are using the musl builds? (PyOxidizer?)

src/validation.rs
```diff
@@ -935,6 +934,14 @@ fn validate_elf<Elf: FileHeader<Endian = Endianness>>(
...
     ));
 
+    // On musl, we don't use `$ORIGIN`
+    if target_triple.contains("-musl") {
+        allowed_libraries.push(format!("libpython{}.so.1.0", python_major_minor));
+        allowed_libraries.push(format!("libpython{}d.so.1.0", python_major_minor));
+        allowed_libraries.push(format!("libpython{}t.so.1.0", python_major_minor));
+        allowed_libraries.push(format!("libpython{}td.so.1.0", python_major_minor));
+    }
+
```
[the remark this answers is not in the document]
zanieb: Oh thanks the changes around 924 are confusing? I'll revert that.

I can also probably make this bit conditional on the `static` build option too.

cpython-unix/build-xextproto.sh
```diff
@@ -15,10 +15,21 @@ export PKG_CONFIG_PATH=/tools/deps/share/pkgconfig
...
+    EXTRA_CONFIGURE_FLAGS="cross_compiling=yes"
+    fi
+fi
+
 CFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC" CPPFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC" LDFLAGS="${EXTRA_TARGET_LDFLAGS}" ./configure \
     --build=${BUILD_TRIPLE} \
     --host=${TARGET_TRIPLE} \
-    --prefix=/tools/deps
+    --prefix=/tools/deps \
+    "${EXTRA_CONFIGURE_FLAGS}"
```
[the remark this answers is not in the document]
zanieb: https://github.com/astral-sh/python-build-standalone/pull/541/commits/a745841d5b641e729d2bf83fcafa7c612b7da2a3

src/validation.rs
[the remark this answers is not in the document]
zanieb: https://github.com/astral-sh/python-build-standalone/pull/541/commits/6a79ece8abbc6f6390eac99625161333e34ac3d3

zanieb:
> I'm assuming we were previously implicitly doing this for musl?

I'm not sure why I can't reply to this one, but yes. We were normalizing the triple to `gnu`.

zanieb:
> Are there any other high-profile users we should check, especially ones we know are using the musl builds? (PyOxidizer?)

I want to ack this, but I don't have a good answer. @ofek comes to mind as an automated consumer outside Astral.

ofek:
Changing the default to something that actually works is desirable in my opinion. It doesn't matter if there is new naming because in my projects that consume (PyApp & Hatch) there is no ability to use newer non-custom builds without a new release because PyApp has a [script](https://github.com/ofek/pyapp/blob/v0.26.0/scripts/update_distributions.py) that generates the hardcoded [source of truth](https://github.com/ofek/pyapp/blob/v0.26.0/build.rs#L29), which Hatch then uses in a [script](https://github.com/pypa/hatch/blob/hatch-v1.14.0/scripts/update_distributions.py) to hardcode its [source of truth](https://github.com/pypa/hatch/blob/hatch-v1.14.0/src/hatch/python/distributions.py). So I would just have to change the former script.

zanieb:
I considered a transition plan like

- Continue to publish static builds as we are today
- Create a duplicate archive with `+static`
- Publish dynamic builds with `+shared`
- Switch the default to the `+shared` variant, retaining a `+shared` duplicate archive
- Drop the `+shared` archive

However, users still need to encode a specific version of python-build-standalone in which the default changed. After thinking about it more, I'm still leaning towards just changing the defaut.

geofft (approved):

geofft:
@ofek by "actually works" I assume you mean "able to dynamically load stuff"?

Then I think we should ship this and call it a bugfix and make sure we ship it at a time that people will be around on Discord etc. to help downstream folks that are broken by the change if any are actually broken by it.

We could also figure out how to make a static musl include a working libdl, which is probably easier than it would be for glibc because musl is generally less complex, but would still be a little bit of a research project.

zanieb:
The diff is not too complicated, so I'll just merge this and close #546

PR #538, opened 2025-02-25
Set `enable-experimental-jit=yes-off` on Python 3.13 / Linux

zanieb:
See #535 

This builds the JIT, but disables it by default. Users can opt-in to enable it at runtime.

3.14 and macOS support will follow, there are some hiccups there.

PR #544, opened 2025-02-27
Allow enabling the JIT on aarch64 macOS

zanieb:
Follows #538


zanieb:
Huh, this passed locally but fails here. Need to dig deeper, I guess.

PR #545, opened 2025-02-27
Populate CXX during builds to avoid calling the system toolchain

### LINKED ISSUE #512: Possibly non-isolated build on macOS
Hi, thanks for maintaining this wonderful project :)

If I understood correctly, one of this project's goals is to isolate as good as possible for builds on macOS, which is why a fixed version of LLVM is pulled in during build time.

I wanted to try making specific builds for the various Apple M processors by adding e.g. `-mcpu=apple-m3` to the `target_cflags` and `target_ldflags` in the `aarch64-apple-darwin` target in `targets.yml`. (Btw, is that a cursed thing to try?)

For `-mcpu=apple-m2` everything works nicely, but for `-mcpu=apple-m3` (or m4) the build fails. I couldn't find anything informative in the logs that would point to the issue, so maybe someone can reproduce this?

But that kind of fits a suspicion I have, namely that the LLVM from XCode is notably involved in the build process (which it shouldn't?) and it does not know about M3 or M4 processors, which is why the build fails. When I run `clang --version` I get `Apple clang version 16.0.0 (clang-1600.0.26.3)` and I'm using macOS `15.1.1 (24B91)` on a M3 MacBook so the LLVM version indeed does not know about M3 and M4. See https://releases.llvm.org/ and https://en.wikipedia.org/wiki/Apple_M3 for the according dates.

Don't know how relevant this is to you but maybe this points to some unintended non-isolation.

zanieb:
Closes #512

### DIFF
--- a/cpython-unix/build.py
+++ b/cpython-unix/build.py
@@ -101,6 +101,9 @@ def add_target_env(env, build_platform, target_triple, build_env):
     env["HOST_CC"] = settings["host_cc"]
     env["HOST_CXX"] = settings["host_cxx"]
     env["CC"] = settings["target_cc"]
+    # We always set CXX, otherwise a build could bypass our toolchain
+    # accidentally, e.g., on macOS where `g++` links to the system clang.
+    env["CXX"] = settings.get("target_cxx", settings.get("host_cxx"))
 
     if settings.get("bolt_capable"):
         env["BOLT_CAPABLE"] = "1"


zanieb:
Concretely, it seems like this means that `libffi` was previously built using `g++` / Apple clang on macOS and now it'll use our LLVM toolchain as intended.

zanieb:
We may just want to define `target_cxx` everywhere instead of falling back to `host_cxx`.

geofft (approved):
Good find!

Mildly unhinged idea - I wonder if we can use the Mac's Seatbelt sandboxing system to allowlist executables from the system to protect against things like this.

Issue #542, opened 2025-02-26
Refactor static and dynamic handling to avoid hard-coding musl


zanieb:
With https://github.com/astral-sh/python-build-standalone/pull/541, musl is switching to dynamic linking and it became apparent that the build system special-cases musl specifically instead of tracking if the build is statically or dynamically linked. It'd be great to add an abstraction to capture the concept rather than checking the target triple.

It's unclear if we'll _need_ this long-term, but it's not a lot of work anyway.

zanieb:
Done in #541

PR #546, opened 2025-02-27
Refactor static / dynamic linking into build options

zanieb:
A bit on the fence about _where_ to put this, i.e., it could be a part of the target triple? Trying this out to start.

The goal here is to untangle the static vs shared builds from musl per #542 so we can ship both kinds in #541

### DIFF
--- a/cpython-unix/build-main.py
+++ b/cpython-unix/build-main.py
@@ -45,6 +45,7 @@ def main():
         print("Unsupported build platform: %s" % sys.platform)
         return 1
 
+    # Note these arguments must be synced with `build.py`
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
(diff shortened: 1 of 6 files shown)


zanieb:
After trying #547 I think this is a little simpler because we otherwise have to sanitize all the targets when building packages, e.g., `<triple>-static` is not a valid `--host` target.

cpython-unix/build-main.py
```diff
@@ -54,10 +55,17 @@ def main():
...
 
-    optimizations = {"debug", "noopt", "pgo", "lto", "pgo+lto"}
+    # Construct possible options, we use a set here for canonical ordering
+    options = set()
+    options.update({"debug", "noopt", "pgo", "lto", "pgo+lto"})
+    options.update({f"freethreaded+{option}" for option in options})
+    link_modes = {"static", "shared"}
+    options.update(
+        {f"{option}+{link_mode}" for link_mode in link_modes for option in options}
+    )
```
zanieb, line 65: I don't like this, especially since it's repeated. I will refactor this further separately.

zanieb:
Some history on shared / static fields can be found by following https://github.com/astral-sh/python-build-standalone/pull/241

zanieb:
Merged via https://github.com/astral-sh/python-build-standalone/pull/541

PR #560, opened 2025-03-15
Pin third-party GitHub Actions

zanieb:
We should probably switch to Renovate's pin support here, but this is an easy first step.


charliermarsh (approved):

charliermarsh:
Did you do this manually?

indygreg (approved):

zanieb:
Yeah I did

PR #562, opened 2025-03-17
Compile sqlite with FTS3 enhanced query syntax enabled

### LINKED ISSUE #550: Compile SQLite with `-DENABLE_FTS3_PARENTHESIS`
We were evaluating using this python distribution for our Docker image instead of the official python3 image, but we've encountered an interesting difference in our test suite: the SQLite extension isn't compiled with the same flags.

<details>
<summary>With docker.io/library/python</summary>

```
$ docker run -it --rm python
Python 3.13.2 (main, Feb 25 2025, 21:31:02) [GCC 12.2.0] on linux
>>> import sqlite3
>>> for (a,) in sqlite3.connect(":memory:").execute("PRAGMA compile_options;").fetchall():
...   print(a)
...
ATOMIC_INTRINSICS=1
COMPILER=gcc-12.2.0
DEFAULT_AUTOVACUUM
DEFAULT_CACHE_SIZE=-2000
DEFAULT_FILE_FORMAT=4
DEFAULT_JOURNAL_SIZE_LIMIT=-1
DEFAULT_MMAP_SIZE=0
DEFAULT_PAGE_SIZE=4096
DEFAULT_PCACHE_INITSZ=20
DEFAULT_RECURSIVE_TRIGGERS
DEFAULT_SECTOR_SIZE=4096
DEFAULT_SYNCHRONOUS=2
DEFAULT_WAL_AUTOCHECKPOINT=1000
DEFAULT_WAL_SYNCHRONOUS=2
DEFAULT_WORKER_THREADS=0
ENABLE_COLUMN_METADATA
ENABLE_DBSTAT_VTAB
ENABLE_FTS3
ENABLE_FTS3_PARENTHESIS
ENABLE_FTS3_TOKENIZER
ENABLE_FTS4
ENABLE_FTS5
ENABLE_LOAD_EXTENSION
ENABLE_MATH_FUNCTIONS
ENABLE_PREUPDATE_HOOK
ENABLE_RTREE
ENABLE_SESSION
ENABLE_STMTVTAB
ENABLE_UNLOCK_NOTIFY
ENABLE_UPDATE_DELETE_LIMIT
HAVE_ISNAN
LIKE_DOESNT_MATCH_BLOBS
MALLOC_SOFT_LIMIT=1024
MAX_ATTACHED=10
MAX_COLUMN=2000
MAX_COMPOUND_SELECT=500
MAX_DEFAULT_PAGE_SIZE=32768
MAX_EXPR_DEPTH=1000
MAX_FUNCTION_ARG=127
MAX_LENGTH=1000000000
MAX_LIKE_PATTERN_LENGTH=50000
MAX_MMAP_SIZE=0x7fff0000
MAX_PAGE_COUNT=1073741823
MAX_PAGE_SIZE=65536
MAX_SCHEMA_RETRY=25
MAX_SQL_LENGTH=1000000000
MAX_TRIGGER_DEPTH=1000
MAX_VARIABLE_NUMBER=250000
MAX_VDBE_OP=250000000
MAX_WORKER_THREADS=8
MUTEX_PTHREADS
OMIT_LOOKASIDE
SECURE_DELETE
SOUNDEX
SYSTEM_MALLOC
TEMP_STORE=1
THREADSAFE=1
USE_URI
>>>
```

</details>

<details>
<summary>With this build</summary>

```
docker run --rm -it ghcr.io/astral-sh/uv:debian uv run --python 3.13 python
Python 3.13.2 (main, Feb 12 2025, 14:38:11) [GCC 6.3.0 20170516] on linux
>>> import sqlite3
>>> for (a,) in sqlite3.connect(":memory:").execute("PRAGMA compile_options;").fetchall():
...     print(a)
...
ATOMIC_INTRINSICS=1
COMPILER=gcc-6.3.0 20170516
DEFAULT_AUTOVACUUM
DEFAULT_CACHE_SIZE=-2000
DEFAULT_FILE_FORMAT=4
DEFAULT_JOURNAL_SIZE_LIMIT=-1
DEFAULT_MMAP_SIZE=0
DEFAULT_PAGE_SIZE=4096
DEFAULT_PCACHE_INITSZ=20
DEFAULT_RECURSIVE_TRIGGERS
DEFAULT_SECTOR_SIZE=4096
DEFAULT_SYNCHRONOUS=2
DEFAULT_WAL_AUTOCHECKPOINT=1000
DEFAULT_WAL_SYNCHRONOUS=2
DEFAULT_WORKER_THREADS=0
DIRECT_OVERFLOW_READ
ENABLE_DBSTAT_VTAB
ENABLE_FTS3
ENABLE_FTS4
ENABLE_FTS5
ENABLE_GEOPOLY
ENABLE_MATH_FUNCTIONS
ENABLE_RTREE
MALLOC_SOFT_LIMIT=1024
MAX_ATTACHED=10
MAX_COLUMN=2000
MAX_COMPOUND_SELECT=500
MAX_DEFAULT_PAGE_SIZE=8192
MAX_EXPR_DEPTH=1000
MAX_FUNCTION_ARG=127
MAX_LENGTH=1000000000
MAX_LIKE_PATTERN_LENGTH=50000
MAX_MMAP_SIZE=0x7fff0000
MAX_PAGE_COUNT=0xfffffffe
MAX_PAGE_SIZE=65536
MAX_SQL_LENGTH=1000000000
MAX_TRIGGER_DEPTH=1000
MAX_VARIABLE_NUMBER=32766
MAX_VDBE_OP=250000000
MAX_WORKER_THREADS=8
MUTEX_PTHREADS
SYSTEM_MALLOC
TEMP_STORE=1
THREADSAFE=1
```

</details>

This has the impact of full-text queries not working properly:

```
$ docker run -it --rm python
Python 3.13.2 (main, Feb 25 2025, 21:31:02) [GCC 12.2.0] on linux
>>> import sqlite3
>>> con = sqlite3.connect(":memory:")
>>> con.execute("CREATE VIRTUAL TABLE docs USING fts4(message);").fetchall()
[]
>>> con.execute("INSERT INTO docs VALUES ('Message number 4');").fetchall()
[]
>>> con.execute("SELECT * FROM docs WHERE docs MATCH 'Message AND 4';").fetchall()
[('Message number 4',)]
```

and with this build:

```
$ docker run --rm -it ghcr.io/astral-sh/uv:debian uv run --python 3.13 python
Python 3.13.2 (main, Feb 12 2025, 14:38:11) [GCC 6.3.0 20170516] on linux
>>> import sqlite3
>>> con = sqlite3.connect(":memory:")
>>> con.execute("CREATE VIRTUAL TABLE docs USING fts4(message);").fetchall()
[]
>>> con.execute("INSERT INTO docs VALUES ('Message number 4');").fetchall()
[]
>>> con.execute("SELECT * FROM docs WHERE docs MATCH 'Message AND 4';").fetchall()
[]
```

FWIW, those are the flags Debian uses for building sqlite (and therefore the sqlite extension in the `docker.io/library/python` image): https://sources.debian.org/src/sqlite3/3.40.1-2%2Bdeb12u1/debian/rules/#L43-L61

So, I'm interested in having the `-DENABLE_FTS3_PARENTHESIS` flag enabled in particular, but it might be worth aligning with how popular distributions compile SQLite?

It's worth noting that [the documentation about FTS in SQLite](https://www.sqlite.org/fts3.html) states:

> New applications should also define the [SQLITE_ENABLE_FTS3_PARENTHESIS](https://www.sqlite.org/compile.html#enable_fts3_parenthesis) macro to enable the [enhanced query syntax](https://www.sqlite.org/fts3.html#_set_operations_using_the_enhanced_query_syntax)

sandhose:
Fixes #550

This adds three compilation flags to the embedded sqlite module:

 - `-DSQLITE_ENABLE_FTS3` to explicitly enable FTS3 support (which AFAIK is already enabled by default)
 - `-DSQLITE_ENABLE_FTS3_PARENTHESIS` to enable the enhanced FTS3 query syntax (which isn't enabled by default)
 - `-DSQLITE_ENABLE_FTS3_TOKENIZER` to enable the two-argument version of the `fts3_tokenizer()` interface. It still has to be enabled on the connection-level using `Connection.setconfig`

All those flags are enabled by default in the Debian build of SQLite, and probably in other distributions as well.


zanieb (approved):
Thank you!

The failures look like #567 — we can ignore those.

geofft:
I know this was a while ago, but

> * `-DSQLITE_ENABLE_FTS3_TOKENIZER` to enable the two-argument version of the `fts3_tokenizer()` interface. It still has to be enabled on the connection-level using `Connection.setconfig`

I don't think this is right - it seems to me that this compile-time flag impacts the _default_ setting of the connection-level flag, and it can still be enabled via `setconfig`, and turning on this flag means it does not need to be enabled via `setconfig`. Compare these versions from before/after this change:

```pycon
$ uvx python@3.12.8
Python 3.12.8 (main, Jan 14 2025, 22:49:14) [Clang 19.1.6 ] on linux
>>> import sqlite3
>>> s = sqlite3.connect(":memory:")
>>> c = s.cursor()
>>> s.getconfig(sqlite3.SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER)
False
>>> c.execute("SELECT fts3_tokenizer('mytokenizer', x'deadbeefdeadbeef')")
  File "<stdin>", line 1, in <module>
sqlite3.OperationalError: fts3tokenize disabled
>>> s.setconfig(sqlite3.SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER)
>>> c.execute("SELECT fts3_tokenizer('mytokenizer', x'deadbeefdeadbeef')")
<sqlite3.Cursor object at 0x78ac33227d40>
>>> ^D
$ uvx python@3.12.9
Python 3.12.9 (main, Mar 17 2025, 21:01:58) [Clang 20.1.0 ] on linux
>>> import sqlite3
>>> s = sqlite3.connect(":memory:")
>>> c = s.cursor()
>>> s.getconfig(sqlite3.SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER)
True
>>> c.execute("SELECT fts3_tokenizer('mytokenizer', x'deadbeefdeadbeef')")
<sqlite3.Cursor object at 0x7d354b927cc0>
```

Are you actually using the two-argument form of `fts3_tokenizer` in your code? (It seems kind of annoying to use from Python, to be honest.) If so, is running `connection.setconfig(sqlite3.SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER)` a workable option for your code? It sounds like this is disabled by default for a relatively good security reason, to prevent a SQL injection from turning into an arbitrary native-code pointer injection. If applications that need it can easily opt in, I'd like to turn it back off by default.

In everyone's defense, the [SQLite documentation](https://www.sqlite.org/fts3.html#custom_application_defined_tokenizers) seems very wrong about this - it claims that SQLite 3.11 continues to enable the two-argument form by default but requires the blob to be a bound parameter instead of a literal, and it doesn't discuss the `#define`. In reality, SQLite 3.11 disables the two-argument form by default, and if you re-enable it (either with the `#define` or with `setconfig`), it doesn't distinguish between bound parameters and literals, as demonstrated above.

I reported the doc bug at https://sqlite.org/forum/forumpost/923cdbe766, and I might raise this to Debian if I get confirmation that I'm understanding this right.

geofft:
OK, I misunderstood this slightly but I think the overall point about leaving the compile-time flag off still applies.

The SQLite docs are correct that the flag only affects whether you must use a bound parameter when calling `fts3_tokenize` or whether it also accepts literals/computed values. As you can see my example code above does not use a bound parameter... if you do use a bound parameter, you can register a tokenizer without the compile-time flag _and_ without the `setconfig` runtime flag.

```pycon
$ uvx python@3.12.8
Python 3.12.8 (main, Jan 14 2025, 22:49:14) [Clang 19.1.6 ] on linux
>>> import sqlite3
>>> s = sqlite3.connect(":memory:")
>>> c = s.cursor()
>>> s.getconfig(sqlite3.SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER)
False
>>> c.execute("SELECT fts3_tokenizer('mytokenizer', ?)", (0xdeadbeefdeadbeef.to_bytes(8, "little"),))
<sqlite3.Cursor object at 0x7e4aac023ec0>
>>> c.execute("SELECT fts3_tokenizer('mytokenizer')").fetchone()
(None,)
>>> c.execute("SELECT fts3_tokenizer('notmytokenizer')").fetchone()
  File "<stdin>", line 1, in <module>
sqlite3.OperationalError: unknown tokenizer: notmytokenizer
```

I would expect that anyone who's registering a tokenizer would _prefer_ to use a bound parameter instead of string interpolation, honestly.

The SQLite docs are incorrect about the effect of the `setconfig` flag (it currently claims that it turns on/off the `fts3_tokenizer` function entirely as opposed to turning off/on the requirement to use bound parameters), which they will fix. They also don't mention the effect of the compile-time flag, which is indeed to just choose the default setting of the `setconfig` flag.

I plan on turning the compile-time flag off in python-build-standalone unless there's a good reason (like existing code) that needs it, and I'll file a new Debian bug about this.

PR #564, opened 2025-03-17
Fix missing `length` in `IntegrityError`

zanieb:
This error is being raised over in https://github.com/astral-sh/python-build-standalone/pull/563 but failing because the length is missing?


zanieb:
I guess this will only fail if there's an intermittent error?

PR #565, opened 2025-03-17
Increase sleep during possible rate limit errors in downloads

zanieb:
We're seeing 403s for bzip2 which suggests we're encountering a rate limit. Let's increase the sleep so we aren't spamming the URL.


geofft (approved):

zanieb:
Caching seems cleaner for now

PR #566, opened 2025-03-17
Cache dependency downloads

zanieb:


geofft (approved):

.github/workflows/linux.yml
```diff
@@ -216,6 +216,15 @@ jobs:
...
           merge-multiple: true
 
+      - name: Cache downloads
+        uses: actions/cache@v4
+        with:
+          path: build/downloads
+          key: ${{ matrix.target_triple }}-${{ hashFiles('pythonbuild/downloads.py')}}
+          restore-keys: |
+            ${{ matrix.target_triple }}-${{ hashFiles('pythonbuild/downloads.py')}}
+            ${{ matrix.target_triple }}-
```
zanieb, line 226: We could also add the OS to the key and use that as a final level match?

Issue #87, opened 2021-08-24
musl-based aarch64 build


achimnol:
I’d like to have the aarch64 version of musl-based build. It seems that I could just add a new triple to the build matrix.
Is there anything more to consider? If it’s that simple, I’d like to make a PR.

indygreg:
I'd like to see this distribution flavor too. However, it might be subtly more challenging than adding a new entry to the YAML. That's because we may have to teach various build scripts to support a cross architecture build **and** musl. Currently, those things are mutually exclusive. I welcome you to try though.

achimnol:
Actually I've tried to build this repository by myself, but encountered some unexpected errors while bootstraping the build toolchain after more than 15 hours of execution. So I just gave up currently as I don't have enough resources to work on. 😭

PR #569, opened 2025-03-19
Add support for aarch64 musl

zanieb:


zanieb:
The failure is

```
m4> checking for gcc... /usr/bin/x86_64-linux-musl-gcc
m4> checking whether the C compiler works...
m4> no
```

which is unsurprising

jakeybrown92:
Do we know when is likely to be released? currently blocking running on latest arm64 on mac and inside our arm64 runner builds 

docker run -it --platform=linux/arm64 ghcr.io/astral-sh/uv:python3.10-alpine uv python install 3.12
error: No download found for request: cpython-3.12-linux-aarch64-musl

zanieb:
@jakeybrown92 not anytime soon, it's non-trivial as mentioned several times (e.g., https://github.com/astral-sh/python-build-standalone/issues/87#issuecomment-927373080).

jakeybrown92:
@zanieb fair enough.. I thought you meant the fix was non trivial. So there isn't currently any workaround when using alpine on latest arm64?

zanieb:
I would use the Python versions from the distribution instead of the uv managed ones for now.

jakeybrown92:
@zanieb okay thanks. We're currently doing that but just wanted to explore UV, but don't really want to change OS. No worries. Appreciate the replies

cpython-unix/build-bzip2.sh
```diff
@@ -15,6 +15,23 @@ else
...
+            if [ -e "${TOOLS_PATH}/host/lib/${filename}" ]; then
+                echo "warning: ${filename} already exists"
+            fi
+            cp "$lib" ${TOOLS_PATH}/host/lib/
+        fi
+    done
+
+    # Add explicit library path for compiler-rt builtins
+    export EXTRA_TARGET_LDFLAGS="${EXTRA_TARGET_LDFLAGS} -L${TOOLS_PATH}/host/lib -lclang_rt.builtins"
+fi
```
zanieb, line 33: This all very dubious

cpython-unix/build.py
```diff
@@ -88,6 +88,11 @@ def add_target_env(env, build_platform, target_triple, build_env):
     extra_host_cflags = []
     extra_host_ldflags = []
 
+    # Add compiler-rt for aarch64-musl to resolve missing builtins
```
zanieb, line 91: Honestly I don't understand why we need to be explicit about this.

Without this, the bzip2 and openssl builds fail with missing symbols (it's possible more do too, I didn't play whackamole for long).

PR #568, opened 2025-03-19
Add support for Python 3.14 on musl

zanieb:
Closes #558
Follow-up to #443

PR #571, opened 2025-03-20
Enable free-threaded builds of Python 3.14 on musl

zanieb:
Extends #568

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -401,6 +401,22 @@ if [ "${CC}" = "musl-clang" ]; then
     done
 fi
 
+# To enable mimalloc (which is hard requirement for free-threaded versions, but preferred in
+# general), we need `stdatomic.h` which is not provided by musl. It's are part of the include files
+# that are part of clang. But musl-clang eliminates them from the default include path. So copy it
+# into place.
+if [[ "${CC}" = "musl-clang" && -n "${PYTHON_MEETS_MINIMUM_VERSION_3_13}" ]]; then
+    for h in /tools/${TOOLCHAIN}/lib/clang/*/include/stdatomic.h; do
+        filename=$(basename "$h")
+        if [ -e "/tools/host/include/${filename}" ]; then
+            echo "${filename} already exists; don't need to copy!"
+            exit 1
+        fi
+        cp "$h" /tools/host/include/
+    done
+fi
+
+
 if [ -n "${CPYTHON_STATIC}" ]; then
     CFLAGS="${CFLAGS} -static"
     CPPFLAGS="${CPPFLAGS} -static"
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -416,7 +416,7 @@ if [ -n "${CPYTHON_DEBUG}" ]; then
 fi
 
 # Explicitly enable mimalloc on 3.13+, it's already included by default but with this it'll fail
-# if it's missing from the system. The musl builds do not supprt mimalloc yet.
+# if it's missing from the system. The MUSL builds do not supprt mimalloc yet.
 if [[ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_13}" && "${CC}" != "musl-clang" ]]; then
     CONFIGURE_FLAGS="${CONFIGURE_FLAGS} --with-mimalloc"
 fi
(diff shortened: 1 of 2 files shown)


zanieb:
```
cpython-3.14> checking for --with-mimalloc...
cpython-3.14> configure: error: mimalloc requires stdatomic.h, use --without-mimalloc to disable mimalloc.
```

It's not clear why this fails yet. I presumed this was a problem with Clang 14, but this should be available there. We should reproduce and look at a `config.log`. 

I was kindly linked to https://gist.github.com/corona10/09459355b64d395aba14eba5be6f6f63 which showed a compiler-rt issue as the culprit (not using our clang version or build system!):

```
configure:18535: checking for stdatomic.h
configure:18558: clang-19 -o conftest   -fsanitize=thread  conftest.c  >&5
/usr/bin/ld: cannot find /usr/local/lib/clang/19/lib/x86_64-unknown-linux-gnu/libclang_rt.tsan.a: No such file or directory
clang-19: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'll setup a local build so I can see if the same thing is happening to us.

cc @indygreg in case there's something obvious here?

indygreg:
Why is thread sanitizer not working a fatal configure error? That seems like the underlying issue here.

zanieb:
I'm not sure. It might not be relevant — that build includes `--with-thread-sanitizer` and ours does not. I'll get logs for our setup.

zanieb:
Yeah that looks irrelevant

```
configure:18506: checking for stdatomic.h
configure:18529: musl-clang -o conftest  -fPIC -I/tools/deps/include -I/tools/deps/include/ncursesw  -fPIC -I/tools/deps/include -I/tools/deps/include/ncursesw  -L/tools/deps/lib -Wl,--exclude-libs,ALL -LModules/_hacl conf
test.c -ldl  -lpthread >&5
conftest.c:158:14: fatal error: 'stdatomic.h' file not found
    #include <stdatomic.h>
             ^~~~~~~~~~~~~
1 error generated.
configure:18529: $? = 1
configure: failed program was:

...

| 
| 
|     #include <stdatomic.h>
|     atomic_int int_var;
|     atomic_uintptr_t uintptr_var;
|     int main() {
|       atomic_store_explicit(&int_var, 5, memory_order_relaxed);
|       atomic_store_explicit(&uintptr_var, 0, memory_order_relaxed);
|       int loaded_value = atomic_load_explicit(&int_var, memory_order_seq_cst);
|       return 0;
|     }
| 
| 
configure:18541: result: no
```

zanieb:
This file is present in Clang 14 (as expected)

```
root@5e42d8cf2e23:~/Python-3.14.0a6# find $(musl-clang -print-file-name=include) -name stdatomic.h
/tools/llvm/lib/clang/14.0.3/include/stdatomic.h
```

zanieb:
Trying https://github.com/astral-sh/python-build-standalone/pull/571/commits/d335fbf8f7f2c9c6f27950adbb289b507ec9c1bb as we do for the intrinsic headers for blake2

indygreg:
It's possible musl-clang's implicit compiler options are not using clang's normal includes search paths. That musl-clang wrapper does weird things.

There's an arg to tell clang to dump all its arguments upon invocation. Can't recall offhand. Add that to see what's actually going on.

cpython-unix/build-cpython.sh
```diff
@@ -407,6 +407,22 @@ if [ "${CC}" = "musl-clang" ]; then
...
+# general), we need `stdatomic.h` which is not provided by musl. It's are part of the include files
+# that are part of clang. But musl-clang eliminates them from the default include path. So copy it
+# into place.
+if [[ "${CC}" = "musl-clang" && -n "${PYTHON_MEETS_MINIMUM_VERSION_3_13}" ]]; then
+    for h in /tools/${TOOLCHAIN}/lib/clang/*/include/stdatomic.h; do
+        filename=$(basename "$h")
+        if [ -e "/tools/host/include/${filename}" ]; then
+            echo "${filename} already exists; don't need to copy!"
+            exit 1
+        fi
```
Gankra, line 420: is this my bash scripting knowledge failing me or are you bailing on the entire build here if the include file exists?

Gankra (approved):

cpython-unix/build-cpython.sh
zanieb: uhh I guess so, I copied this from above though. It seems fine to bail because it's a wrong assumption in the build script.

PR #572, opened 2025-03-20
Add option to keep build contexts

zanieb:
i.e., Docker containers and temporary directories.

I have been using a patch like this locally and added an environment variable so we can just have it available. I rebased this quickly and need to double check it works as intended still.


zanieb:
Oh, today I also found `PYBUILD_BREAK_ON_FAILURE`.. which is fairly compelling

PR #592, opened 2025-04-19
Statically link libpython into interpreter (but keep building libpython3.x.so)

geofft:
Opening PR just to see how far CI gets. I've done some very minimal testing locally (in a CPython tree, not using the p-b-s infrastructure) and it seems to do what I want.

We can probably make some changes to patchelf as a result.

The added patch here is [geofft/cpython branch static-interpreter](https://github.com/python/cpython/compare/main...geofft:cpython:static-interpreter). I'll open a PR to CPython and link it once CI looks good.


geofft:
No obvious performance impact in any direction, really.

In the following:
* directory b contains `cpython-3.13.3-x86_64-unknown-linux-gnu-pgo+lto` from the previous workflow run where I incorrectly had the patch applied after `autoconf` so it didn't do anything (which makes it a perfect baseline for comparison), I think [this workflow run](https://github.com/astral-sh/python-build-standalone/actions/runs/14715224988) specifically
* directory c contains `cpython-3.13.3-x86_64-unknown-linux-gnu-pgo+lto` from the current run
* and then I ran `cp -a c new-build-with-old-binary; cp {b,new-build-with-old-binary}/python/install/bin/python3.13` to get a Python executable that uses libpython.so, i.e, in the new-build-with-old-binary directory, we are loading an unBOLTed libpython into the interpreter, just to see what libpython's performance is like.
* pystone.py is the last version from before it was removed from cpython sources (`git show 61fd70e05027150b21184c7bc9fa8aa0a49f9601^:Lib/test/pystone.py`)
* test machine is an AWS t2.medium running the Ubuntu 24.04 AMI, recently rebooted / not running anything else

```
ubuntu@ip-172-16-0-59:~$ hyperfine -L dir b,c,new-build-with-old-binary '{dir}/python/install/bin/python3 -c "import ssl"'
Benchmark 1: b/python/install/bin/python3 -c "import ssl"
  Time (mean ± σ):      35.2 ms ±   1.3 ms    [User: 27.7 ms, System: 7.3 ms]
  Range (min … max):    33.1 ms …  41.9 ms    70 runs
 
Benchmark 2: c/python/install/bin/python3 -c "import ssl"
  Time (mean ± σ):      35.2 ms ±   0.9 ms    [User: 27.1 ms, System: 8.0 ms]
  Range (min … max):    33.2 ms …  37.8 ms    82 runs
 
Benchmark 3: new-build-with-old-binary/python/install/bin/python3 -c "import ssl"
  Time (mean ± σ):      35.5 ms ±   1.0 ms    [User: 28.4 ms, System: 6.9 ms]
  Range (min … max):    33.6 ms …  38.7 ms    84 runs
 
Summary
  b/python/install/bin/python3 -c "import ssl" ran
    1.00 ± 0.04 times faster than c/python/install/bin/python3 -c "import ssl"
    1.01 ± 0.05 times faster than new-build-with-old-binary/python/install/bin/python3 -c "import ssl"
ubuntu@ip-172-16-0-59:~$ hyperfine -L dir b,c,new-build-with-old-binary '{dir}/python/install/bin/python3 -c "import ssl"'
Benchmark 1: b/python/install/bin/python3 -c "import ssl"
  Time (mean ± σ):      35.6 ms ±   0.8 ms    [User: 28.4 ms, System: 7.0 ms]
  Range (min … max):    33.8 ms …  37.6 ms    81 runs
 
Benchmark 2: c/python/install/bin/python3 -c "import ssl"
  Time (mean ± σ):      35.5 ms ±   0.9 ms    [User: 27.5 ms, System: 7.8 ms]
  Range (min … max):    33.5 ms …  37.8 ms    82 runs
 
Benchmark 3: new-build-with-old-binary/python/install/bin/python3 -c "import ssl"
  Time (mean ± σ):      36.1 ms ±   0.9 ms    [User: 28.5 ms, System: 7.3 ms]
  Range (min … max):    34.1 ms …  38.3 ms    81 runs
 
Summary
  c/python/install/bin/python3 -c "import ssl" ran
    1.00 ± 0.03 times faster than b/python/install/bin/python3 -c "import ssl"
    1.02 ± 0.04 times faster than new-build-with-old-binary/python/install/bin/python3 -c "import ssl"
ubuntu@ip-172-16-0-59:~$ hyperfine -L dir b,c,new-build-with-old-binary '{dir}/python/install/bin/python3 pystone.py'
Benchmark 1: b/python/install/bin/python3 pystone.py
  Time (mean ± σ):     120.3 ms ±   2.3 ms    [User: 114.0 ms, System: 6.0 ms]
  Range (min … max):   116.4 ms … 125.9 ms    25 runs
 
Benchmark 2: c/python/install/bin/python3 pystone.py
  Time (mean ± σ):     121.2 ms ±   1.3 ms    [User: 113.8 ms, System: 7.1 ms]
  Range (min … max):   118.7 ms … 124.5 ms    24 runs
 
Benchmark 3: new-build-with-old-binary/python/install/bin/python3 pystone.py
  Time (mean ± σ):     118.6 ms ±   1.3 ms    [User: 112.8 ms, System: 5.6 ms]
  Range (min … max):   116.5 ms … 120.9 ms    24 runs
 
Summary
  new-build-with-old-binary/python/install/bin/python3 pystone.py ran
    1.01 ± 0.02 times faster than b/python/install/bin/python3 pystone.py
    1.02 ± 0.02 times faster than c/python/install/bin/python3 pystone.py
```

geofft:
We'd also expect to see some speedup on thread-local storage. I can confirm that we are getting the faster TLS model but I can't actually measure a difference.

i've tried doing moar threading with this silly benchmark, but only got a 1-2% performance difference.
```python3
import pystone
import concurrent.futures
t = concurrent.futures.ThreadPoolExecutor()
for i in range(500):
    t.submit(pystone.pystones, 1000)
t.shutdown()
```

*However*, on the free-threaded variants (`cpython-3.13-x86_64-unknown-linux-gnu-freethreaded+pgo+lto`), there is a noticeable performance difference:

```
ubuntu@ip-172-16-0-59:~/ft$ hyperfine -L dir b,c,new-build-with-old-binary '{dir}/python/install/bin/python3 ~/pystone.py'
Benchmark 1: b/python/install/bin/python3 ~/pystone.py
  Time (mean ± σ):     252.2 ms ±   2.5 ms    [User: 243.9 ms, System: 7.9 ms]
  Range (min … max):   248.9 ms … 255.8 ms    12 runs
 
Benchmark 2: c/python/install/bin/python3 ~/pystone.py
  Time (mean ± σ):     227.8 ms ±   1.7 ms    [User: 218.9 ms, System: 8.6 ms]
  Range (min … max):   224.0 ms … 230.0 ms    13 runs
 
Benchmark 3: new-build-with-old-binary/python/install/bin/python3 ~/pystone.py
  Time (mean ± σ):     241.7 ms ±   3.0 ms    [User: 233.0 ms, System: 8.4 ms]
  Range (min … max):   237.1 ms … 248.4 ms    12 runs
 
Summary
  c/python/install/bin/python3 ~/pystone.py ran
    1.06 ± 0.02 times faster than new-build-with-old-binary/python/install/bin/python3 ~/pystone.py
    1.11 ± 0.01 times faster than b/python/install/bin/python3 ~/pystone.py
ubuntu@ip-172-16-0-59:~/ft$ hyperfine -L dir b,c,new-build-with-old-binary '{dir}/python/install/bin/python3 ~/threadstone.py'
Benchmark 1: b/python/install/bin/python3 ~/threadstone.py
  Time (mean ± σ):      2.194 s ±  0.055 s    [User: 3.725 s, System: 0.548 s]
  Range (min … max):    2.111 s …  2.269 s    10 runs
 
Benchmark 2: c/python/install/bin/python3 ~/threadstone.py
  Time (mean ± σ):      1.992 s ±  0.037 s    [User: 3.396 s, System: 0.480 s]
  Range (min … max):    1.924 s …  2.037 s    10 runs
 
Benchmark 3: new-build-with-old-binary/python/install/bin/python3 ~/threadstone.py
  Time (mean ± σ):      2.109 s ±  0.037 s    [User: 3.571 s, System: 0.535 s]
  Range (min … max):    2.051 s …  2.167 s    10 runs
 
Summary
  c/python/install/bin/python3 ~/threadstone.py ran
    1.06 ± 0.03 times faster than new-build-with-old-binary/python/install/bin/python3 ~/threadstone.py
    1.10 ± 0.03 times faster than b/python/install/bin/python3 ~/threadstone.py
```

So I think the change is defensible on those grounds.

(And the non-BOLT libpython isn't slower, and is actually apparently _faster_, which is maybe concerning....)


<details>
<summary>Details of thread-local storage code generation differences:</summary>

From `objdump -dr python/install/lib/libpython3.13.so.1.0`, somewhere in the function `PyPegen_new_identifier`:

```gas
  1ac233:	e8 98 be ff ff       	call   1a80d0 <PyUnicode_DecodeUTF8@plt>
			1ac234: R_X86_64_PLT32	PyUnicode_DecodeUTF8-0x4
  1ac238:	48 89 44 24 08       	mov    %rax,0x8(%rsp)
  1ac23d:	48 85 c0             	test   %rax,%rax
  1ac240:	0f 84 79 01 00 00    	je     1ac3bf <_PyPegen_new_identifier+0x1af>
  1ac246:	f6 40 20 40          	testb  $0x40,0x20(%rax)
  1ac24a:	74 4c                	je     1ac298 <_PyPegen_new_identifier+0x88>
  1ac24c:	48 8d 3d 6d 71 4c 01 	lea    0x14c716d(%rip),%rdi        # 16733c0 <.got>
			1ac24f: R_X86_64_TLSLD	_Py_tss_tstate-0x4
  1ac253:	e8 08 eb ff ff       	call   1aad60 <__tls_get_addr@plt>
			1ac254: R_X86_64_PLT32	__tls_get_addr@GLIBC_2.3-0x4
  1ac258:	48 8b 80 18 00 00 00 	mov    0x18(%rax),%rax
			1ac25b: R_X86_64_DTPOFF32	_Py_tss_tstate
  1ac25f:	48 8b 78 10          	mov    0x10(%rax),%rdi
  1ac263:	4c 8d 74 24 08       	lea    0x8(%rsp),%r14
  1ac268:	4c 89 f6             	mov    %r14,%rsi
  1ac26b:	e8 d0 d2 ff ff       	call   1a9540 <_PyUnicode_InternImmortal@plt>
			1ac26c: R_X86_64_PLT32	_PyUnicode_InternImmortal-0x4
  1ac270:	48 8b 7b 20          	mov    0x20(%rbx),%rdi
  1ac274:	4d 8b 36             	mov    (%r14),%r14
  1ac277:	4c 89 f6             	mov    %r14,%rsi
  1ac27a:	e8 31 bb ff ff       	call   1a7db0 <_PyArena_AddPyObject@plt>
			1ac27b: R_X86_64_PLT32	_PyArena_AddPyObject-0x4
```

From `objdump -dr python/install/bin/python3`, same code:

```gas
 1c7bdb3:	e8 9e 39 dc ff       	call   1a3f756 <PyUnicode_DecodeUTF8>
 1c7bdb8:	48 89 44 24 08       	mov    %rax,0x8(%rsp)
 1c7bdbd:	48 85 c0             	test   %rax,%rax
 1c7bdc0:	0f 84 d6 d0 2c 00    	je     1f48e9c <_PyPegen_new_identifier.cold+0x108>
 1c7bdc6:	f6 40 20 40          	testb  $0x40,0x20(%rax)
 1c7bdca:	0f 84 c4 cf 2c 00    	je     1f48d94 <_PyPegen_new_identifier.cold>
 1c7bdd0:	64 48 8b 04 25 f8 ff 	mov    %fs:0xfffffffffffffff8,%rax
 1c7bdd7:	ff ff 
 1c7bdd9:	48 8b 78 10          	mov    0x10(%rax),%rdi
 1c7bddd:	4c 8d 74 24 08       	lea    0x8(%rsp),%r14
 1c7bde2:	4c 89 f6             	mov    %r14,%rsi
 1c7bde5:	e8 e0 32 dc ff       	call   1a3f0ca <_PyUnicode_InternImmortal>
 1c7bdea:	48 8b 7b 20          	mov    0x20(%rbx),%rdi
 1c7bdee:	4d 8b 36             	mov    (%r14),%r14
 1c7bdf1:	4c 89 f6             	mov    %r14,%rsi
 1c7bdf4:	e8 e3 1a dc ff       	call   1a3d8dc <_PyArena_AddPyObject>
```

Note the function call to `__tls_get_addr` in the shared library and the direct reference to 8 bytes below `%fs` in the binary.

</details>

cpython-unix/build-cpython.sh
```diff
@@ -298,6 +298,20 @@ if [ -n "${PYTHON_MEETS_MAXIMUM_VERSION_3_10}" ]; then
...
+    patch -p1 -i "${ROOT}/patch-python-configure-add-enable-static-libpython-for-interpreter.patch"
+else
+    patch -p1 -i "${ROOT}/patch-python-configure-add-enable-static-libpython-for-interpreter-${PYTHON_MAJMIN_VERSION}.patch"
+fi
+
+# When cross-compiling, ./configure checks that run code cannot be run,
+# and some of the defaults are wrong or unwanted, so we have to fix
+# those manually.
+# TODO(geofft): This cannot be the only thing.
+patch -p1 -i "${ROOT}/patch-python-configure-cross-assume-pthread.patch"
```
zanieb, line 313: Can we pass an option to `configure` that set this to yes instead of patching?

zanieb: (Should this be limited to cross-compiles?)

zanieb (approved):

cpython-unix/build-cpython.sh
geofft: > (Should this be limited to cross-compiles?)

The patch as written is limited to cross-compiles, in that it changes the default behavior when ./configure is running in cross-compile mode and therefore (thinks that) it cannot run test programs. It leaves the configure test in place with the current behavior for the non-cross-compiling case. See #599.

> Can we pass an option to `configure` that set this to yes instead of patching?

Looks like yes, with the test ./configure script mentioned in #599:
```sh-session
$ ./configure cross_compiling=yes ac_cv_pthread=yes
[...]
checking whether gcc accepts -pthread... (cached) yes
```

My vague intuition is that a patch is the right thing here because this patch is actually suitable for upstream—it is a better guess in 2025 that a compiler supports `-pthread` than that it doesn't (just as it's a better guess that a compiler supports C99 `strftime`, etc.). But I don't have a strong opinion. If you prefer, yes, we can do
```sh
if [ -n "${CROSS_COMPILING}" ]; then
  EXTRA_CONFIGURE_FLAGS="${EXTRA_CONFIGURE_FLAGS} ac_cv_pthread=yes"
fi
```
(As mentioned in #599 we can also pass `cross_compiling=yes` on the command line and drop the existing patch-force-cross-compile.patch.)

geofft:
Yeah, I think this needs a small tweak to the validation script, let me do that. I'm going to tag this PR as riscv only to avoid rerunning CI on the successful arches, for future reference [here's the current Actions run with those results](https://github.com/astral-sh/python-build-standalone/actions/runs/14868870698).

cpython-unix/build-cpython.sh
zanieb: I do think
```
if [ -n "${CROSS_COMPILING}" ]; then
  EXTRA_CONFIGURE_FLAGS="${EXTRA_CONFIGURE_FLAGS} ac_cv_pthread=yes"
fi
```
is generally better than a patch, since it removes a layer of indirection and is a common `./configure` pattern. It also makes it much more obvious that it's a cross-compile only change. I'm also supportive of you trying to contribute the patch upstream. I don't feel strongly here, but I was surprised / confused.

zanieb:
Nice! Those last few fixes look good to me.

PR #595, opened 2025-04-25
Patch `uuid.getnode()` to use a different MAC address identification method

zanieb:
Takes https://github.com/python/cpython/pull/132901

Needs to be backported to 3.13, etc.


zanieb:
This doesn't seem to work on aarch64 Linux?

```
❯ docker run -it -v ./python:/python ubuntu:latest /bin/bash
root@20b506c09fbd:/# /python/install/bin/python -c "import uuid; print(uuid.getnode())"
276090243745164
root@20b506c09fbd:/# /python/install/bin/python -c "import uuid; print(uuid.getnode())"
148934309301581
```

PR #598, opened 2025-04-29
Docs: remove Windows static build quirk

### LINKED ISSUE #597: Documentation: Clarify state of python static builds on Windows
The current docs say that [Windows Static Distributions are Extremely Brittle](https://github.com/astral-sh/python-build-standalone/blob/main/docs/quirks.rst#windows-static-distributions-are-extremely-brittle).
However, what are these static builds? For example, there is https://github.com/astral-sh/python-build-standalone/issues/221, basically suggesting that static Windows builds are no longer a thing, concluded with #234.
This suggests to me that the docs are not up to date at this point. Any improvement of the docs would be greatly appreciated here.

LeonarddeR:
The docs say that [Windows Static Distributions are Extremely Brittle](https://github.com/astral-sh/python-build-standalone/blob/main/docs/quirks.rst#windows-static-distributions-are-extremely-brittle).
However, since #234, Windows static builds no longer exist, making this documentation obsolete.

Fixes #597

### DIFF
--- a/docs/quirks.rst
+++ b/docs/quirks.rst
@@ -109,33 +109,6 @@ To use pip, run ``python.exe -m pip``. (It is generally a best practice to
 invoke pip via ``python -m pip`` on all platforms so you can be explicit
 about the ``python`` executable that pip uses.)
 
-.. _quirk_windows_static_distributions:
-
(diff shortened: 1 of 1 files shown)


zanieb (approved):
Thanks!

Issue #576, opened 2025-04-03
how to run setcap on python executable


jsiverskog:
we're having some special requirements. we're using the [pysoem](https://github.com/bnjmnp/pysoem) library to communicate with ethercat devices. this means that the python executable needs `cap_net_raw+ep` capability, which can be achieved by running `sudo setcap cap_net_raw+ep /path/to/python`.

this works fine with the system python, but when setting it on the `uv` provided python executable:
```
sudo setcap cap_net_raw+ep ~/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11
```
i get this when i then try to run python:
```
[...]/python3: error while loading shared libraries: $ORIGIN/../lib/libpython3.11.so.1.0: DST not allowed in SUID/SGID programs
```

if i run this:
```
patchelf --replace-needed "\$ORIGIN/../lib/libpython3.11.so.1.0" ~/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/libpython3.11.so.1.0  ~/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11
```
it works (after re-running `setcap`), so it seems to be related to the relative path. and sure, from a security point of view i can understand why that may be problematic.

i found this: https://github.com/astral-sh/python-build-standalone/blob/f0abfc9cb1f6a985fc5561cf5435f7f6e8a64e5b/cpython-unix/build-cpython.sh#L684-L697

which is probably what causes this. but what is the right forward here? running pop!_os 22.04 lts (based on ubuntu 24.04) amd64

indygreg:
Apparently the glibc dynamic loader straight up refuses dynamic token expansion in DT_NEEDED for setuid/setgid binaries.

That error message and logic originates from a commit in 1999!

geofft:
Yeah, this is defensible on glibc's part because historically you can hardlink to a binary you don't own, so some other user on the machine could make a hardlink in a directory they control, meaning that `$ORIGIN/../lib/` would have their own libraries which could run with the added privileges. (On modern Linux systems this is disabled with /proc/sys/fs/restricted_hardlinks, but I am not sure if that's enough to justify glibc turning off this protection.)

Agree re statically linking libpython into the python3 binary - I'll ping here once I have something for you to try.

---

[...]

PR #621, opened 2025-05-22
Add an rpath to bin/python3 on glibc platforms to fix #619

### LINKED ISSUE #619: Latest changes to statically link `libpython` breaks usage of packages like `python-mscl`
Hello!

I noticed that after upgrading `uv` to 0.7.6 and reinstalling python (`uv python install --reinstall`), the import of a package like `python-mscl` breaks. Here's a traceback:

```
Python 3.13.3 (main, May 17 2025, 13:34:03) [GCC 6.3.0 20170516] on linux
>>> from python_mscl import mscl
  File "<python-input-0>", line 1, in <module>
    from python_mscl import mscl
  File "/home/pi/AirbrakesV2/.venv/lib/python3.13/site-packages/python_mscl/__init__.py", line 1, in <module>
    from . import mscl
  File "/home/pi/AirbrakesV2/.venv/lib/python3.13/site-packages/python_mscl/mscl.py", line 13, in <module>
    from . import _mscl
ImportError: libpython3.13.so.1.0: cannot open shared object file: No such file or directory
```
----
Since I'm also the maintainer of `python-mscl`, I can give some more insight on how that package works:

My [package](https://github.com/harshil21/python-mscl) is just a wrapper around `https://github.com/LORD-MicroStrain/MSCL`, aiming to simplify the installation procedure of the library, which is a little convoluted and not easy to setup on new machines - [link](https://github.com/LORD-MicroStrain/MSCL/blob/master/HowToUseMSCL.md#python-1). 

So I am essentially just bundling the `.so` and `.py` files they provide into a package for every python version and architecture. And this just works on import, since it can find the .so and .py files in the same directory.

----

So I'm not sure if there's something wrong with how I'm packaging this library, and I have to correct it somehow, or if this was a problem with how my python was built. Since this happened immediately after reinstalling my python, and I saw the release note which mentioned this change, I thought I should open an issue to seek some clarification.

geofft:
Even though the Python interpreter no longer needs libpython3.x.so, it turns out some extension modules (incorrectly) do, and miraculously, allowing them to find libpython3.x.so doesn't actually break things, for reasons detailed in the comment. So set an rpath that allows libpython3.x.so to be loaded if needed by some other library, even though we won't use that ourselves. We are already doing this on musl; do it on glibc too.

Note that this change does not risk users who want to make bin/python3 setuid or setcap (e.g. #576); while the rpath is presumably ignored for privileged binaries, there is no error message, and the binary launches fine, and _because_ we do not need the rpath in order for the interpreter to work, everything (except these misbuilt extension modules) works.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -688,44 +688,104 @@ if [ "${PYBUILD_SHARED}" = "1" ]; then
                 -change /install/lib/${LIBPYTHON_SHARED_LIBRARY_BASENAME} @executable_path/../lib/${LIBPYTHON_SHARED_LIBRARY_BASENAME} \
                 ${ROOT}/out/python/install/bin/python${PYTHON_MAJMIN_VERSION}${PYTHON_BINARY_SUFFIX}
         fi
-    else
+    else # (not macos)
         LIBPYTHON_SHARED_LIBRARY_BASENAME=libpython${PYTHON_MAJMIN_VERSION}${PYTHON_BINARY_SUFFIX}.so.1.0
         LIBPYTHON_SHARED_LIBRARY=${ROOT}/out/python/install/lib/${LIBPYTHON_SHARED_LIBRARY_BASENAME}
 
-        if [ "${CC}" == "musl-clang" ]; then
-            # musl does not support $ORIGIN in DT_NEEDED, so we use RPATH instead. This could be
-            # problematic, i.e., we could load the shared library from the wrong location if
-            # `LD_LIBRARY_PATH` is set, but there's not a clear alternative at this time. The
-            # long term solution is probably to statically link to libpython instead.
-            patchelf --set-rpath "\$ORIGIN/../lib" \
-                ${ROOT}/out/python/install/bin/python${PYTHON_MAJMIN_VERSION}
+        # Although we are statically linking libpython, some extension
+        # modules link against libpython.so even though they are not
+        # supposed to do that. If you try to import them on an
+        # interpreter statically linking libpython, all the symbols they
+        # need are resolved from the main program (because neither glibc
+        # nor musl has two-level namespaces), so there is hopefully no
+        # correctness risk, but they need to be able to successfully
+        # find libpython.so in order to load the module.  To allow such
+        # extensions to load, we set an rpath to point at our lib
+        # directory, so that if anyone ever tries to find a libpython,
+        # they successfully find one. See
+        # https://github.com/astral-sh/python-build-standalone/issues/619
+        # for some reports of extensions that need this workaround.
+        #
+        # Note that this matches the behavior of Debian/Ubuntu/etc.'s
+        # interpreter (if package libpython3.x is installed, which it
+        # usually is thanks to gdb, vim, etc.), because libpython is in
+        # the system lib directory, as well as the behavior in practice
+        # on conda-forge miniconda and probably other Conda-family
+        # Python distributions, which too set an rpath.
+        #
+        # There is a downside of making this libpython locatable: some user
+        # code might do e.g.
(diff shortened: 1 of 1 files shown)


cpython-unix/build-cpython.sh
```diff
@@ -688,45 +688,105 @@ if [ "${PYBUILD_SHARED}" = "1" ]; then
...
+        # they successfully find one. See
+        # https://github.com/astral-sh/python-build-standalone/issues/619
+        # for some reports of extensions that need this workaround.
+        #
+        # Note that this matches the behavior of Debian/Ubuntu/etc.'s
+        # interpreter (if package libpython3.x is installed, which it
+        # usually is thanks to gdb, vim, etc.), because libpython is in
+        # the system lib directory, as well as the behavior in practice
+        # on conda-forge miniconda and probably other Conda-family
+        # Python distributions, which too set an rpath.
```
zanieb, line 714: typo? costly to rebuild :D

cpython-unix/build-cpython.sh
```diff
@@ -688,45 +688,105 @@ if [ "${PYBUILD_SHARED}" = "1" ]; then
...
+    else # (not macos)
         LIBPYTHON_SHARED_LIBRARY_BASENAME=libpython${PYTHON_MAJMIN_VERSION}${PYTHON_BINARY_SUFFIX}.so.1.0
         LIBPYTHON_SHARED_LIBRARY=${ROOT}/out/python/install/lib/${LIBPYTHON_SHARED_LIBRARY_BASENAME}
 
-        if [ "${CC}" == "musl-clang" ]; then
-            # musl does not support $ORIGIN in DT_NEEDED, so we use RPATH instead. This could be
-            # problematic, i.e., we could load the shared library from the wrong location if
-            # `LD_LIBRARY_PATH` is set, but there's not a clear alternative at this time. The
-            # long term solution is probably to statically link to libpython instead.
+        # Although we are statically linking libpython, some extension
```
zanieb, line 695: The column wrap on this is pretty intense given the amount of content and the presence of longer lines in the file, but 🤷‍♀️ 

Overall, this feels a little more long-form than I'd personally do here when you could just have a brief version and put the rest it in the pull request, but I don't feel strongly about the styling here.

cpython-unix/build-cpython.sh
```diff
@@ -688,45 +688,105 @@ if [ "${PYBUILD_SHARED}" = "1" ]; then
...
-                ${ROOT}/out/python/install/bin/python${PYTHON_MAJMIN_VERSION}
+                ${ROOT}/out/python/install/bin/python${PYTHON_MAJMIN_VERSION}${PYTHON_BINARY_SUFFIX}
+        fi
 
+        # For libpython3.so (the ABI3 library for embedders), we do
+        # still dynamically link libpython3.x.so.1.0 (the
+        # version-specific library), because there is no particular
+        # speedup/benefit in statically linking libpython into
+        # libpython3.so, and we'd just be shipping a third copy of the
+        # libpython code. Therefore we use the old logic for that and
```
zanieb, line 769: "old logic" here feels confusing in a comment, old compared to when? It seems like this will go stale.

zanieb (approved):

cpython-unix/build-cpython.sh
geofft: don't think so, or I'm not seeing it? "which, also, set an rpath."

Thanks for the style feedback, I may or may not roll it into an update of this file depending on how builds go :) My overall instinct is that there's enough going on that I want things like this highly visible to the next person to think "can I change this" but yeah I could have done that differently.

(Also, oops, should have tagged this linux, doing that now before I forget just in case we have to rebuild)

zanieb: Ah I see. I got confused by the phrasing haha. Definitely understand following up separately.

PR #622, opened 2025-05-23
Disable unsafe identical code folding in BOLT

geofft:
astral-sh/uv#13610 reported a misbehavior that is the result of a subclass of str incorrectly having its ->tp_as_number->nb_add slot filled in with the value of PyUnicode_Type->tp_as_sequence->sq_concat. There are some times when this is an appropriate thing to do iwhen subclassing, but this is not one of them. The logic to prevent it in this case relies on two helper functions in the file, wrap_binaryfunc and wrap_binaryfunc_l, having different addresses, even though they contain identical code.

For some reason BOLT does not do this optimization in the shared library (even though those are static functions and not exported), so we only started seeing this in the static build.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -271,6 +271,12 @@ if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_12}" ]; then
     # https://github.com/python/cpython/issues/128514
     patch -p1 -i ${ROOT}/patch-configure-bolt-apply-flags-128514.patch
 
+    # Disable unsafe identical code folding. Objects/typeobject.c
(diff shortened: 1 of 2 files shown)


geofft:
We very much need to upstream this change, btw, but let's get this shipped.

andersk:
Related:

- https://github.com/python/cpython/issues/53093#issuecomment-1093504696
- https://github.com/python/cpython/issues/101525

geofft:
Ah, thanks for the links, Anders! list + tuple subclass and list + str subclass presumably trigger the same behavior, yeah. And indeed that second link also confirms that this only happens with static libpython.

I opened python/cpython#134642 which is a more thorough version of this change. (It doesn't matter for us since we know we're on LLVM 20, but I'll update the patch in this PR next week.)

zanieb (approved):

PR #624, opened 2025-05-23
Go back to dynamically linking libpython for now

geofft:


geofft:
Cancelling - we used an older release in uv instead (astral-sh/uv#13617).

PR #625, opened 2025-05-26
Update quirks doc to match current reality

geofft:
We configure ncurses with defaults for TERMINFO_DIRS that work on I can find (it works on Debian family, Red Hat family, NixOS, and Alpine), so we shouldn't phrase either the title or the body as if this is a problem that affects most people or setting the environment variable is a necessary thing in general. (Also I think even in cases where we don't have terminfo for your terminal, backspace usually still works, as evidenced by `TERM=potato python`).

musl distributions are now dynamic binaries by default and dynamic loading works.

The libcrypt transition is behind us.

The python3-config file (since its first version, I think) knows how to correct the prefix in output. While it does contain hard-coded paths internally, it uses that for the purpose of fixing its output, so I don't think that fact is helpful to call out.

python-build-standalone users who get it through uv are (hopefully) not affected by absolute paths in sysconfig; call this out explicitly for the benefit of people evaluting whether uv will work for them. Also mention sysconfigpatcher for everyone else.

In general, tighten up the content.

### DIFF
--- a/docs/quirks.rst
+++ b/docs/quirks.rst
@@ -4,17 +4,21 @@
 Behavior Quirks
 ===============
 
+While these Python distributions are intended to be broadly compatible
+with the Python ecosystem, there are a few known behavior quirks that
+affect specific environments, packages, or use cases.
+
 .. _quirk_backspace_key:
 
-Backspace Key Doesn't work in Python REPL
-=========================================
+If special keys do not work in the Python REPL
+==============================================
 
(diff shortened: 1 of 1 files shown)


geofft:
I keep forgetting that rST has no provision for putting a link inside bold text, and maybe the next thing I'll do is switch to Markdown in some fashion (MyST if necessary but I think we don't even need that).

zanieb (approved):

PR #626, opened 2025-05-26
Remove support for Linux x86 builds

indygreg:
We stopped shipping the distributions in 6fa8fe9f42aa9187095d60a8dacf19e4eb76baf6 in October 2023. At the time we preserved build support just in case.

I don't believe anyone has complained about the loss of x86 Linux distributions.

I don't believe there is compelling value in PBS supporting x86 Linux distributions.

If someone wants to run Python on x86 Linux, they can always use CPython's official build systen, which still supports targeting x86 Linux.

This commit rips out our support building for x86 Linux.


geofft (approved):
OK with me. Is there a particular reason you want to take it out or just general cleanup?

indygreg:
General cleanup. I want to do some housekeeping on the container images as part of standing up native aarch64 builds.

PR #627, opened 2025-05-26
Use host CPython as Python dependency; remove xcb image

indygreg:
The `xcb` Dockerfile / image existed to provide a host build of CPython, which was required to build libxcb, xcb-proto, and tk. The existence of this image predated our building a host CPython in order to build CPython itself.

With PBS now unconditionally building a host CPython, we can use our host CPython during the builds of these packages, eliminating the need for the xcb image.

This commit makes that transition.

We also upgrade libxcb and xcb-proto to the latest version as part of this migration. Older versions of libxcb aren't compatible with Python 3.12+ due to removal of the `imp` module.


geofft (approved):
Great, the `xcb` image confused me :)

PR #635, opened 2025-05-30
Use `gh release` instead of someone's release action

geofft:
The upstream action times out because it requests every release (without pagination), and there's no real need to use a third-party action for this when the `gh` command does a perfectly fine job too.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
       - run: just release-dry-run ${{ secrets.GITHUB_TOKEN }} ${{ github.event.inputs.sha }} ${{ github.event.inputs.tag }}
         if: ${{ github.event.inputs.dry-run == 'true' }}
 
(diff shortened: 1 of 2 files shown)


zanieb (approved):
Nice, I hate Actions!

geofft:
Meant to send this earlier - I am pretty sure this matches the behavior of the third-party action as we were using it. (I don't know whether we actually want the behavior of reusing a previous prerelease... given that you have to manually delete the tag anyway, it doesn't really make anything easier, and I kind of don't want assets uploaded that the most recent run wouldn't have uploaded.)

I am not totally sure how to test this. I could some releases with 1980s version numbers and see what happens? I could make a detached fork in my account and try it out? Alternatively I'll likely do the next release and I can just make sure it does what I expect, if we're comfortable merging without testing.

zanieb:
I'm fine with moving forward without testing for now.

PR #648, opened 2025-06-05
fix(workflow): use correct input variable for release-create step

wojiushixiaobai:
The release-create step was using an incorrect variable name (`github.events.input.tag`), which resulted in the `just release-create` command being called without the required tag argument. This commit updates the workflow to use the correct variable (`github.event.inputs.tag`), ensuring the tag is properly passed to the command.

- https://github.com/loong64/python-build-standalone/actions/runs/15469452121/job/43549720438

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,7 +66,7 @@ jobs:
       # Create a GitHub release.
       - name: Create GitHub Release
         if: ${{ github.event.inputs.dry-run == 'false' }}
-        run: just release-create ${{ github.events.input.tag }}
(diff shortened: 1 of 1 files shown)


geofft:
Thanks, sorry about that! I had this in 60711c436744050b74031c278eb60f1b628145d6 which I used for yesterday's release, but I forgot to open a pull request, so I'll merge yours. :)

BTW I will take a look at your question about loongarch support soon - I think I would like to merge support into this repo so you no longer have to maintain a fork (and deal with any bugs in our release process :) ).

PR #663, opened 2025-06-17
Bump to CPython 3.14.0b3

zanieb:


zanieb:
Reporting the JIT failure upstream in https://github.com/python/cpython/issues/135690

PR #607, opened 2025-05-12
Enable native builds on Linux aarch64

indygreg:
I just published an LLVM 20 toolchain for aarch64. The toolchain has support for PGO and BOLT.

This commit switches the Linux aarch64 builds to be performed natively on aarch64 machines. PGO and BOLT are enabled on the builds, hopefully making them a bit faster.

PR #670, opened 2025-06-28
Enable native builds on Linux aarch64

zanieb:
- **xtrans 1.5.0 -> 1.6.0**
- **x11-util-macros 1.20.1 -> 1.20.2**
- **xorgproto 2019.1 -> 2024.1**
- **Enable native builds on Linux aarch64**


zanieb:
As I mentioned at https://github.com/astral-sh/python-build-standalone/pull/607#issuecomment-3015988434 — I need to review this myself as I did not review the original implementation and there may be some clean-up to be done after hacking things to work.

cpython-unix/build-main.py
```diff
@@ -95,10 +95,12 @@ def main():
             "toolchain-image-build",
             "toolchain-image-build.cross",
             "toolchain-image-build.cross-riscv64",
+            "toolchain-image-build.debian9",
             "toolchain-image-gcc",
             "toolchain-image-xcb",
             "toolchain-image-xcb.cross",
             "toolchain-image-xcb.cross-riscv64",
+            "toolchain-image-xcb.debian9",
```
zanieb, line 103: See https://github.com/astral-sh/python-build-standalone/pull/670/files#r2173842155

cpython-unix/build.py
```diff
@@ -496,10 +504,14 @@ def python_build_info(
 
     binary_suffix = ""
 
-    if platform == "linux_x86_64":
+    if platform in ("linux_x86_64", "linux_aarch64"):
+        arch = "aarch64" if platform == "linux_aarch64" else "x86_64"
```
zanieb, line 508: We should probably make this exhaustive

ci-matrix.py
```diff
@@ -25,7 +25,10 @@
     {"name": "build", "arch": "x86_64"},
     {"name": "build.cross", "arch": "x86_64"},
     {"name": "build.cross-riscv64", "arch": "x86_64"},
+    {"name": "build.debian9", "arch": "aarch64"},
```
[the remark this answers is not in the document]
zanieb: > In general - is "debian9" here because we're mostly using Debian 8 and that's too old for aarch64?

I presume so, that's where I got stuck at https://github.com/astral-sh/python-build-standalone/pull/484#issuecomment-2741233269 but this was Greg's choice.

Issue #95, opened 2021-10-12
PyQt segfaults on modern Linux distros when run with standalone python


dae:
This is a curious one. Something about the way these packages are built is causing a segfault when PyQt tries to initialize a connection to the X server. The following can be seen on distros like Ubuntu 21.04, Debian 11 or Fedora 35:

```shell
% tar -axf cpython-3.9.6-x86_64-unknown-linux-gnu-pgo-20210724T1424.tar.zst
% ./python/install/bin/python3 -m venv{,-test} && . ./venv-test/bin/activate && pip install pyqt5 && DISPLAY=:0 python -c 'import sys; from PyQt5.QtGui import QGuiApplication as Q; Q(sys.argv)'; deactivate
Collecting pyqt5
  Using cached PyQt5-5.15.4-cp36.cp37.cp38.cp39-abi3-manylinux2014_x86_64.whl (8.3 MB)
Collecting PyQt5-Qt5>=5.15
  Using cached PyQt5_Qt5-5.15.2-py3-none-manylinux2014_x86_64.whl (59.9 MB)
Collecting PyQt5-sip<13,>=12.8
  Using cached PyQt5_sip-12.9.0-cp39-cp39-manylinux1_x86_64.whl (328 kB)
Installing collected packages: PyQt5-Qt5, PyQt5-sip, pyqt5
Successfully installed PyQt5-Qt5-5.15.2 PyQt5-sip-12.9.0 pyqt5-5.15.4
WARNING: You are using pip version 20.2.3; however, version 21.3 is available.
You should consider upgrading via the '/home/dae/venv-test/bin/python3 -m pip install --upgrade pip' command.
Segmentation fault (core dumped)
```

If we run it with the system-installed Python 3.9 or one built from standard Python sources, it works fine:

```shell
% python3.9 -m venv{,-test} && . ./venv-test/bin/activate && pip install pyqt5 && DISPLAY=:0 python -c 'import sys; from PyQt5.QtGui import QGuiApplication as Q; Q(sys.argv)'; deactivate
Collecting pyqt5
  Using cached PyQt5-5.15.4-cp36.cp37.cp38.cp39-abi3-manylinux2014_x86_64.whl (8.3 MB)
Collecting PyQt5-Qt5>=5.15
  Using cached PyQt5_Qt5-5.15.2-py3-none-manylinux2014_x86_64.whl (59.9 MB)
Collecting PyQt5-sip<13,>=12.8
[...]

PR #676, opened 2025-06-29
Build libtcl, libtk, and _tkinter as shared objects, and remove Tix

### LINKED ISSUE #129: Can't import matplotlib on linux: module '_tkinter' has no attribute '__file__'
Basic matplotlib usage such as `import matplotlib.pyplot as plt; plt.plot(list(range(10)), list(range(10)))` crashes with `module '_tkinter' has no attribute '__file__'`. 

On system python, _tkinter is a shared library (`/usr/lib/python3.8/lib-dynload/_tkinter.cpython-38-x86_64-linux-gnu.so`), while in these python builds it's a builtin module which does not have `__file__`. matplotlib expects tkinter to be an .so it can load: 
https://github.com/matplotlib/matplotlib/blob/f25c2d025fa8f70d6688e7acf4b85e4eb7590299/src/_tkagg.cpp#L310-L322

To reproduce:

```bash
wget https://github.com/indygreg/python-build-standalone/releases/download/20220502/cpython-3.10.4+20220502-x86_64_v3-unknown-linux-gnu-pgo+lto-full.tar.zst
tar xf cpython-3.10.4+20220502-x86_64_v3-unknown-linux-gnu-pgo+lto-full.tar.zst
python/install/bin/python3 -m venv venv
venv/bin/pip install matplotlib
venv/bin/python -c "import matplotlib.pyplot as plt; plt.plot(list(range(10)), list(range(10)))"
```

```
  File "<string>", line 1, in <module>
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 2769, in plot
    return gca().plot(
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 2274, in gca
    return gcf().gca(**kwargs)
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 867, in gcf
    return figure()
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 808, in figure
    manager = new_figure_manager(
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 326, in new_figure_manager
    _warn_if_gui_out_of_main_thread()
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 316, in _warn_if_gui_out_of_main_thread
    if (_get_required_interactive_framework(_get_backend_mod())
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 217, in _get_backend_mod
    switch_backend(dict.__getitem__(rcParams, "backend"))
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 268, in switch_backend
    switch_backend(candidate)
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 288, in switch_backend
    class backend_mod(matplotlib.backend_bases._Backend):
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/pyplot.py", line 289, in backend_mod
    locals().update(vars(importlib.import_module(backend_name)))
  File "/home/konsti/monotrail/pbsi/python/install/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/backends/backend_tkagg.py", line 1, in <module>
    from . import _backend_tk
  File "/home/konsti/monotrail/pbsi/venv/lib/python3.10/site-packages/matplotlib/backends/_backend_tk.py", line 24, in <module>
    from . import _tkagg
AttributeError: module '_tkinter' has no attribute '__file__'. Did you mean: '__name__'?
```

Running `TCL_LIBRARY="$(pwd)/python/install/lib/tcl8.6" venv/bin/python -c "import matplotlib.pyplot as plt; plt.plot(list(range(10)), list(range(10)))"` instead as suggested by the docs makes no difference.

EDIT: I'm mainly using the standalone builds by libloading `libpython.so` and launching one of the very high level c api commands, so i'd be happy about any workaround that works with libloading or importlib hackery from rust.

### LINKED ISSUE #146: Not able to run a tkinter program
I am using the bazel/rules_python to compile our python code. 
according to the configuration, we are using the 
cpython-3.9.13+20220802-x86_64-unknown-linux-gnu-install_only.tar.gz as our python interpreter.
But as we are using the Tkinter, there is a segment fault during the runtime.
the code
```python
import faulthandler
faulthandler.enable()

import tkinter
from tkinter.constants import *
tk = tkinter.Tk()
frame = tkinter.Frame(tk, relief=RIDGE, borderwidth=2)
frame.pack(fill=BOTH,expand=1)
label = tkinter.Label(frame, text="Hello, World")
label.pack(fill=X, expand=1)
button = tkinter.Button(frame,text="Exit",command=tk.destroy)
button.pack(side=BOTTOM)
tk.mainloop()
```

the stack trace
```
/tmp/python_interpreter_x86_64-unknown-linux-gnu ❯ ./bin/python3.9 tkinter_test.py                                                                                                                                                                                            
Fatal Python error: Segmentation fault

Current thread 0x00007fe29108ef40 (most recent call first):
  File "/tmp/python_interpreter_x86_64-unknown-linux-gnu/lib/python3.9/tkinter/__init__.py", line 2572 in __init__
  File "/tmp/python_interpreter_x86_64-unknown-linux-gnu/lib/python3.9/tkinter/__init__.py", line 3148 in __init__
  File "/tmp/python_interpreter_x86_64-unknown-linux-gnu/tkinter_test.py", line 9 in <module>
[1]    55139 segmentation fault (core dumped)  ./bin/python3.9 tkinter_test.py
```

according to other tickets, seems the python interpreter is already linked to the Tkinter. 

# Update
I was trying to use the debug version to know what happened in the core dump. but it is very strange that I am able to run this with python3.9d. it is very weird.

### LINKED ISSUE #533: Unable to use Pillow ImageTk module
On python-build-standalone Python 3.12 or 3.13 (installed via uv), I tried this with the [Pillow](https://pypi.org/project/pillow/) library (10.4.0, 11.0.0, 11.1.0; wheels or built from source):

```py
import PIL._imagingtk
```

and got the following error:

```
  File "<stdin>", line 1, in <module>
TypeError: bad argument type for built-in operation
```

If that import fails, the `PIL.ImageTk` module doesn't work properly. `_imagingtk` is a compiled C module.

Observed on both macOS and Linux (Arch). The import works fine on system / brew installations of Python.

geofft:
Several important third-party packages, including matplotlib in its
tkagg backend and Pillow, use tkinter as a way of locating libtcl and
libtk and making direct C API calls to those libraries. For more 
details, see the analysis in
https://github.com/astral-sh/python-build-standalone/issues/129#issuecomment-3016267363

To make these packages work, we need to expose the full libtcl and
libtk dynamic symbol ABI; we can't just statically link them into our
own binary. It seems most robust to also expose these as separate
libraries under their usual filenames to match the behavior of other
Python distributions.

Build shared libraries for the _tkinter module and for libtcl and libtk,
and set up rpaths so we find our copies of them. libX11 continues to be
statically linked, but it's linked into libtk. Just as with the build of
Python itself, use --exclude-libs=ALL to prevent the dependencies'
symbols from being exported.

Stop building Tix because it's broken (#723) and it would need to be
changed to dynamic linking.

Configure libX11 with --disable-loadable-xcursor to fix #146, which I
ran into while running tests.

Add zlib as a build-dep of Tcl/Tk so that they can statically link
libz.a. I think we were previously picking up the zlib headers from the
OS, which wasn't a problem when libtcl and libtk were static libraries -
they got linked into CPython itself which also linked zlib.a. But now
libtcl.so and libtk.so need zlib.a.

Fixes #129
Fixes #533

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -173,7 +173,7 @@ jobs:
         run: |
           uv run ci-matrix.py \
             --platform linux \
-            --labels '${STEPS_GET_LABELS_OUTPUTS_LABELS}' \
+            --labels "${STEPS_GET_LABELS_OUTPUTS_LABELS}" \
             --max-shards 2 \
             ${{ (steps.check-pythonbuild.outputs.changed == 'true' || github.ref == 'refs/heads/main') && '--force-crate-build' || '' }} \
             > matrix.json
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -90,7 +90,7 @@ jobs:
       - name: Generate build matrix
         id: set-matrix
         run: |
-          uv run ci-matrix.py --platform darwin --labels '${STEPS_GET_LABELS_OUTPUTS_LABELS}' ${{ (steps.check-pythonbuild.outputs.changed == 'true' || github.ref == 'refs/heads/main') && '--force-crate-build' || '' }} > matrix.json
+          uv run ci-matrix.py --platform darwin --labels "${STEPS_GET_LABELS_OUTPUTS_LABELS}" ${{ (steps.check-pythonbuild.outputs.changed == 'true' || github.ref == 'refs/heads/main') && '--force-crate-build' || '' }} > matrix.json
 
           # Extract python-build matrix
           echo "matrix=$(jq -c '."python-build"' matrix.json)" >> $GITHUB_OUTPUT
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -90,7 +90,7 @@ jobs:
       - name: Generate build matrix
         id: set-matrix
         run: |
-          uv run ci-matrix.py --platform windows --labels '${STEPS_GET_LABELS_OUTPUTS_LABELS}' ${{ (steps.check-pythonbuild.outputs.changed == 'true' || github.ref == 'refs/heads/main') && '--force-crate-build' || '' }} > matrix.json
+          uv run ci-matrix.py --platform windows --labels "${STEPS_GET_LABELS_OUTPUTS_LABELS}" ${{ (steps.check-pythonbuild.outputs.changed == 'true' || github.ref == 'refs/heads/main') && '--force-crate-build' || '' }} > matrix.json
 
           # Extract python-build matrix
           echo "matrix=$(jq -c '."python-build"' matrix.json)" >> $GITHUB_OUTPUT
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "pythonbuild"
 version = "0.1.0"
 authors = ["Gregory Szorc <gregory.szorc@gmail.com>"]
-edition = "2021"
+edition = "2024"
 
 [dependencies]
 anyhow = "1.0.80"
--- a/cpython-unix/Makefile
+++ b/cpython-unix/Makefile
@@ -195,16 +195,6 @@ $(OUTDIR)/sqlite-$(SQLITE_VERSION)-$(PACKAGE_SUFFIX).tar: $(PYTHON_DEP_DEPENDS) 
 $(OUTDIR)/tcl-$(TCL_VERSION)-$(PACKAGE_SUFFIX).tar: $(PYTHON_DEP_DEPENDS) $(HERE)/build-tcl.sh
 	$(RUN_BUILD) --docker-image $(DOCKER_IMAGE_BUILD) tcl
 
-TIX_DEPENDS = \
-    $(HERE)/build-tix.sh \
-    $(OUTDIR)/tcl-$(TCL_VERSION)-$(PACKAGE_SUFFIX).tar \
-    $(OUTDIR)/tk-$(TK_VERSION)-$(PACKAGE_SUFFIX).tar \
-    $(if $(NEED_LIBX11),$(OUTDIR)/libX11-$(LIBX11_VERSION)-$(PACKAGE_SUFFIX).tar) \
-    $(NULL)
-
-$(OUTDIR)/tix-$(TIX_VERSION)-$(PACKAGE_SUFFIX).tar: $(TIX_DEPENDS)
-	$(RUN_BUILD) --docker-image $(DOCKER_IMAGE_BUILD) tix
-
 TK_DEPENDS = \
     $(HOST_PYTHON_DEPENDS) \
     $(HERE)/build-tk.sh \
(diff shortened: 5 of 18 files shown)


geofft:
I think this is ready to review. Seems to work in mild manual testing on macOS and glibc. The formatting check and the failing static musl builds should be quick to fix, I'll do that tomorrow. I want to do a little more extensive testing, including on a dynamic musl distribution, but I expect to be able to undraft this tomorrow hopefully.

Note this is split up into commits and might be easier to review commit-by-commit.

geofft:
> The title should be updated to indicate this removes Tix too.

I think I was thinking of doing rebase-and-merge, which would preserve the separate commit removing it and linking to #723, but I can update the title and squash-and-merge.

zanieb:
Rebase-merge is fine. It seems nice to update the title regardless, but I don't have strong feelings.

johnthagen:
@geofft Does this PR fix or change any of the discussion in the docs at:

https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html#static-linking-of-libx11-incompatibility-with-pyqt-on-linux

I've had issues where `python-build-standalone` apps using Tk didn't work on e.g., Ubuntu 24.04 due to (I think) static linking of X11. Was curious if this PR would substantively change that use case and if that section of the docs should be updated at all?

geofft:
@johnthagen I believe that issue was #95, fixed in May 2022, and so that entire section of the docs has not been relevant for a few years. #739 removes it that section from the docs. If I'm wrong about that / if you have seen this failure more recently, please comment there or open a new bug. :)

It's possible the `--disable-loadable-xcursor` change in this PR also fixes something that could cause similar misbehavior.

johnthagen:
@geofft Thanks. I'll try it again with the latest `python-build-standalone` again. And, I should have mentioned, this was specifically when we were building a PyInstaller executable using `python-build-standalone` as the target for an application that used Tk internally. After the executable was build and deployed to an Ubuntu 24.04 host, there were libx11 issues. It's possible that the issue is an incompatibility/assumption between PyInstaller and `python-build-standalone`. I can try to reproduce the issue and file it for you if it still exists.

johnthagen:
@geofft The issues we were seeing with `python-build-standalone` + Tk + Pyinstaller seems to be fixed in the latest `python-buildstandalone` 🎉 .

For posterity, the issue we were seeing was:

```
../../src/xcb_io.c:157: append_pending_request: Assertion `!xcb_xlib_unknown_seq_number' failed.
Aborted (core dumped)
```

whenever our PyInstaller, `python-build-standalone` Tk binary launched on Ubuntu 24.04/22.04. This seems to be resolved now. The relevant versions:

- CPython 3.12.11
- pyinstaller 6.14.2

PR #690, opened 2025-07-08
Only upload one release artifact at a time

zanieb:
Debugging HTTP 422 release failure https://github.com/astral-sh/python-build-standalone/actions/runs/16156480423


indygreg:
lol GitHub.

Can we work around this with retries instead?

zanieb:
We don't need this, I was just trying to see _which_ file we were getting a 422 on by avoiding concurrency. It turns out the file limit is the issue :/

PR #691, opened 2025-07-09
Drop `.sha256` files from releases

zanieb:
GitHub has started enforcing a 1000 file limit, which breaks us.

I first encountered it in CI during the release: https://github.com/astral-sh/python-build-standalone/actions/runs/16156480423

Then got to an actual error message via the CLI:

```
❯ gh release upload 20250708 dist/cpython-3.14.0b4-riscv64-unknown-linux-gnu-freethreaded+lto-20250708T1536.tar.zst
HTTP 422: Validation Failed (https://uploads.github.com/repos/astral-sh/python-build-standalone/releases/230978017/assets?label=&name=cpython-3.14.0b4-riscv64-unknown-linux-gnu-freethreaded%2Blto-20250708T1536.tar.zst)
file_count limited to 1000 assets per release
```

You can see they documented a limit recently:

https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas

https://web.archive.org/web/20250629152715/https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -505,14 +505,6 @@ pub async fn command_upload_release_distributions(args: &ArgMatches) -> Result<(
                 file_data,
                 dry_run,
             ));
-            fs.push(upload_release_artifact(
-                &raw_client,
-                token.clone(),
-                &release,
-                format!("{}.sha256", dest),
-                Bytes::copy_from_slice(format!("{}\n", digest).as_bytes()),
-                dry_run,
-            ));
         }
 
         let mut buffered = futures::stream::iter(fs).buffer_unordered(16);


zanieb:
I verified this resolves the problem by releasing from this branch https://github.com/astral-sh/python-build-standalone/releases/tag/20250708

Unfortunately I think we'll need to move to the consolidated file without warning, since we can't otherwise publish. Even if we did multiple GitHub Releases, that'd be a breaking change for consumers.

I think @geofft had expressed interest in moving away from the individual checksum files.

geofft:
I'm okay with dropping them and I think the value of it is minimal since it's over HTTPS and from the same web server anyway. Even with 500 artifacts the full SHA256SUMS file is manageable for people who need it.

I'd also like, at some point, to move to our own API that we control, even if it sends redirects to GitHub artifact storage for actual downloads, so things like this aren't forced breaking changes.

Issue #692, opened 2025-07-09
Build issue, file or directory not found


kwmlodozeniec:
When I try to build python locally (using docker, which seems the default), I get the following error and the build fails.
```
FileNotFoundError: [Errno 2] No such file or directory: '/MY_HOME_DIR/python-build-standalone/build/image-gcc.linux_x86_64'
```

I build in WSL2, but I have also tried building on a "real" Linux host and got the same result. This used to work for me in the past, so I'm at a loss as to what has happened.

Full log:

```
./build-linux.py --python cpython-3.10
Collecting attrs==24.3.0 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 3))
  Using cached attrs-24.3.0-py3-none-any.whl (63 kB)
Collecting certifi==2024.12.14 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 9))
  Using cached certifi-2024.12.14-py3-none-any.whl (164 kB)
Collecting charset-normalizer==3.4.1 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 13))
  Using cached charset_normalizer-3.4.1-cp313-cp313-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (144 kB)
Collecting docker==7.1.0 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 107))
  Using cached docker-7.1.0-py3-none-any.whl (147 kB)
Collecting idna==3.10 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 111))
  Using cached idna-3.10-py3-none-any.whl (70 kB)
Collecting jinja2==3.1.5 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 115))
  Using cached jinja2-3.1.5-py3-none-any.whl (134 kB)
Collecting jsonschema==4.23.0 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 119))
  Using cached jsonschema-4.23.0-py3-none-any.whl (88 kB)
Collecting jsonschema-specifications==2024.10.1 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 123))
  Using cached jsonschema_specifications-2024.10.1-py3-none-any.whl (18 kB)
Collecting markupsafe==3.0.2 (from -r /MY_HOME_DIR/python-build-standalone/requirements.txt (line 127))
[...]

PR #693, opened 2025-07-09
Build images when the image id cannot be determined from file

zanieb:
Closes https://github.com/astral-sh/python-build-standalone/issues/692


geofft:
#695 was the right fix for this and this doesn't seem like what we want given https://github.com/astral-sh/python-build-standalone/issues/692#issuecomment-3052513592.

Issue #698, opened 2025-07-11
Run test_tkinter in CI


geofft:
From src/verify_distribution.py:
```python
    @unittest.skipIf("TCL_LIBRARY" not in os.environ, "TCL_LIBRARY not set")
    @unittest.skipIf("DISPLAY" not in os.environ, "DISPLAY not set")
    def test_tkinter(self):
        import tkinter as tk
        ...
```

We ought to be able to run this in Xvfb and ensure it runs, though we'd have to automate clicking the button, either by injecting a button click or by modifying the test to quit on its own.

(It looks like [Tk is pretty heavily tied to X11 at the moment](https://discuss.python.org/t/feature-request-wayland-support-for-tkinter/75198), but if we feel like being complete, we could also test under Xwayland + [some headless Wayland compositor](https://unix.stackexchange.com/q/653672) like `WLR_BACKENDS=headless sway` or `weston --backend=headless`.)

Issue #697, opened 2025-07-11
Run test_curses_interactive in CI


geofft:
From src/verify_distribution.py:
```python
    @unittest.skipIf(os.name == "nt", "curses not available on Windows")
    @unittest.skipIf("TERM" not in os.environ, "TERM not set")
    def test_curses_interactive(self):
        import curses
        
        curses.initscr()
        curses.endwin()
```

I believe this is getting skipped in GitHub Actions, but it's probably enough to run this inside a pty, we just need to set `TERM` to some valid value. If not we can try to run it in a real headless terminal like `screen`/`tmux` or something (just make sure we report the exit status correctly).

PR #699, opened 2025-07-11
verify_distribution: Stop setting TERMINFO_DIRS and TCL_LIBRARY

geofft:
In #215 we started setting the standard terminfo directories by default and in #421 we taught Tcl to find its own install directory without setting environment variables. Stop setting these environment variables in tests to more closely match how users will use our builds.

### DIFF
--- a/src/verify_distribution.py
+++ b/src/verify_distribution.py
@@ -6,36 +6,6 @@ import os
 import sys
 import unittest
 
-TERMINFO_DIRS = [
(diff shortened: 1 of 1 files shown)


geofft:
Neither of these tests run in CI yet (#697, #698) so I'm holding off until either we do that or I get a chance to run the tests by hand, but I wanted to open this PR before I forgot about it.

PR #700, opened 2025-07-11
Skip sqlite3 geopoly test on Windows

geofft:
Also reorder the CI validate/upload steps so that if this goes wrong again we can still cut a release without rebuilding.


geofft:
Reverting this in #800.

PR #716, opened 2025-07-31
chore(ci): apply security fixes, add zizmor workflow

woodruffw:
90% of this is from `--fix=unsafe`, with the rest
being being my manual fixups.

In addition to the fixes themselves, I've done the following:

* I added a new top-level workflow that will flag any additional findings in the future (both inline in PRs and also in GitHub's "advanced security" integration). That latter one might need to be disabled though if the org doesn't have it enabled, though.
* I added GitHub Actions to the Dependabot config, and also enabled [cooldown](https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/) everywhere to make it a bit harder for us to introduce newly broken or compromised deps. It's not perfect, but it's better than realtime consumption given that we're already on a monthly deps schedule 🙂 

Signed-off-by: William Woodruff <william@astral.sh>

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,4 +1,4 @@
-name: GitHub Actions Security Analysis with zizmor 🌈
+name: zizmor
 
 on:
   push:
@@ -10,7 +10,7 @@ permissions: {}
 
 jobs:
   zizmor:
-    name: Run zizmor 🌈
+    name: Run zizmor
     runs-on: ubuntu-latest
     permissions:
       security-events: write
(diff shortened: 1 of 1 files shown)


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,24 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
```
zanieb, line 1: Screenshot 2025-07-31 at 1 46 54 PM
This feels pretty out of place with our usual styling

woodruffw: Yeah, that was copy paste from my template. I'll match the style here.

woodruffw: (Waiting to push to see how far the builds get.)

zanieb (approved):

Issue #724, opened 2025-08-06
Multiarch reporting error in <= 3.12 musl builds + compiled packages broken in 3.11 and 3.12


nathanscain:
## Description

> Combining into 1 bug report as they were found at the same time and seem related - but feel free to split up.

 - python-build-standlone musl builds for 3.9-3.12 report as multiarch of `x86_64-linux-gnu` instead of `x86_64-linux-musl` within `sys.implementation`. 3.13 and 3.14 report the correct multiarch.
- Additionally, 3.11 and 3.12 musl builds are unable to load compiled dependencies. 3.13+ along with 3.9 and 3.10 can use compiled packages like lxml, cryptography, pydantic-core, and numpy; however, 3.11 and 3.12 always fail to load the modules.

I expect that before 3.11 python was using a different mechanism to identify which libc modules to load and it starts working again in 3.13+ when the multiarch setting is fixed.

Note that this is not an issue at all within the official alpine docker images - they all report the correct multiarch and load modules correctly even when the base system exhibits the issue.

## Potential Fix

Looks like there was a previously identified issue with how clang prints something with `--print-multiarch` that confuses the CPython configure. It is being applied to all gnu builds, but only musl builds 3.13+ which lines up with the reporting I am seeing. The relevant patches are 4 years old now, so I expect that a more recent update to clang has made this issue affect musl like it does on gnu and the patch needs to be applied to both.

https://github.com/astral-sh/python-build-standalone/blob/55f3d50aee9743ea48afaa18d43ff7a96af7b056/cpython-unix/build-cpython.sh#L116C1-L130C3

```diff
diff --git a/cpython-unix/build-cpython.sh b/cpython-unix/build-cpython.sh
index cc93f0a..3c252a5 100755
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -116,17 +116,12 @@ fi
 # Clang 13 actually prints something with --print-multiarch, confusing CPython's
 # configure. This is reported as https://bugs.python.org/issue45405. We nerf the
 # check since we know what we're doing.
-if [ "${CC}" = "clang" ]; then
+if [[ "${CC}" = "clang" || "${CC}" = "musl-clang" ]]; then
     if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_13}" ]; then
         patch -p1 -i ${ROOT}/patch-disable-multiarch-13.patch
     else
[...]

PR #725, opened 2025-08-06
apply disable-multiarch patch to all musl builds

nathanscain:
Potential solve for #724


geofft:
This seems intuitively right to me, I can't see a rationale why we wouldn't want this for musl-clang + Python <3.13. Thanks!

nathanscain:
My thought exactly - happy to help!

PR #730, opened 2025-08-07
Skip invalid archives from GitHub during release

zanieb:
It looks like https://github.com/astral-sh/python-build-standalone/pull/705 somehow includes files we don't want now?

See https://github.com/astral-sh/python-build-standalone/actions/runs/16804083046/job/47592503471

https://github.com/astral-sh/python-build-standalone/pull/732 also resolves this, but maybe we want both? We validate we have the files we need later anyway, so it's annoying for this to be fatal.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -54,7 +54,7 @@ async fn fetch_artifact(
     org: &str,
     repo: &str,
     artifact: WorkflowListArtifact,
-) -> Result<bytes::Bytes> {
+) -> Result<(String, bytes::Bytes)> {
     println!("downloading artifact {}", artifact.name);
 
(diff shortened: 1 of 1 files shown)


geofft:
I feel like I'm slightly more inclined towards #732 than this, silently skipping a file looks like a good way to confuse ourselves in the future.

zanieb:
It's not silent! We do log it. It's unlikely there's a consequence here since we validate we have all the artifacts we want? We'll just skip other files "silently" downstream anyway. It seems like without this, the release workflow will be more brittle — it will unexpectedly break when we add some additional intermediate artifact.

PR #731, opened 2025-08-07
Fix label subsetting

zanieb:
Regressed in #716


woodruffw (approved):
Thanks! Sorry for not caching these.

PR #732, opened 2025-08-07
Add more artifacts to the skip list

zanieb:
I noticed we were fetching these but... we shouldn't be :)


geofft:
These are also super noisy in the check summary page e.g. https://github.com/astral-sh/python-build-standalone/actions/runs/16805284921?pr=732 - I'd like to stop generating these if we can, but ignoring them in the release process definitely sounds right.

PR #739, opened 2025-08-08
docs: Move libX11 / PyQt stuff to "former quirks"

geofft:
This has been working fine since May 2022....


geofft:
Honestly we should think about getting rid of the quirks page entirely - the remaining quirks are
* if you're on an unusual OS, we won't be able to find your terminfo
* We don't ship Tix, which is deprecated in all our versions and gone on Python 3.13.
* We don't ship a `pip.exe` on Windows and you have to  use `python -m pip` (or, like, uv)
* If you're using `libpython.a` (which isn't in the install-only tarball) on macOS, it's a little complicated
* We use libedit and other people use readline (I think this is barely noticeable with pyrepl)
* Sometimes you will see build-time paths in places that are hopefully harmless

Issue #742, opened 2025-08-09
Set rpath on our shared libraries, and use ORIGIN-relative dependencies for libtcl and libtk


geofft:
As a result of #619 I added a `DT_RPATH` to bin/python3 to support third-party wheels that (incorrectly) link libpython3.x.so.1.0 and expect to be able to find it. The symbols come from bin/python3, we just needed libraries to be able to not fail when resolving dependencies by filename. In other words, we weren't actually using the `DT_RPATH` ourselves. By the same logic, I didn't add an rpath to libpython3.x.so.1.0 itself, because it didn't need to load any _other_ libraries.

In #676 I added two more shared libraries, libtcl8.6.so and libtk8.6.so. On Linux these are referenced with an unqualified name in _tkinter, and so they only work because of the `DT_RPATH` on bin/python3. That is a bug, both because we shouldn't be relying on the bin/python rpath ourselves (so we have the option to remove it later), and because it means that embedders linking libpython3 won't be able to load _tkinter properly without setting an rpath of their own. (On macOS, these are referenced through an rpath in _tkinter, so I think this works properly for embedders.)

Also, pyinstaller/pyinstaller#9204 requests that we make `ldd .../_tkinter.cpython-xxx.so` report the path of libtcl and libtk properly.

I think I want to make the following changes:
* Set `DT_RPATH` on all of our shared libraries, for consistency. Probably including `libpython3.so` (the ABI3 interface).
* On glibc, use `${ORIGIN}`-relative `DT_NEEDED` entries for the dependenci from tkinter to libtcl and libtk, so that it can still find them if the embedder is using `DT_RUNPATH`, which suppresses `DT_RPATH`.
* Add validation tests that we have rpaths on all dynamic objects and`${ORIGIN}`-relative `DT_NEEDED` entries on glibc except for things that are provided by the OS.
* Analogously on macOS, ensure we have relative `LC_LOAD_DYLIB` commands for everything not provided by the OS. (This can be `@executable_path`-relative, `@loader_path`-relative, or `@rpath`-relative if we also have an `LC_RPATH` that is executable- or loader-relative. Note that rpaths work a little differently with macOS dyld - they're not used by default, they're only used for things that are specifically loaded from `@rpath/something.dylib`.)
* Add integration tests for embedders, including an embedder that uses `DT_RUNPATH`.
* Add a specific regression test for `ldd _tkinter` for PyInstaller.

PR #745, opened 2025-08-11
Give _tkinter $ORIGIN-relative dependencies on glibc and an rpath on musl

geofft:
Partially addresses #742 and makes it consistent with what we're doing on macOS.

There's an argument in the comments above that we should not set an rpath on libpython (except on musl where it's needed), and I need to see if I still believe that. In the meantime I'm following that pattern and setting $ORIGIN-relative NEEDED on glibc and rpath on musl only.

This also adds a specific ldd regression test but not the additional tests listed in #742.


geofft:
Giving up on testing this on musl for reasons noted in the comment, but the code to handle musl is there....

LordAro:
Any word on this? I've run into it as well after upgrading from 3.12 to 3.13. I suspect it's only working "incidentally" at the moment as most systems already have tcl installed on the system. But not my dockerised test machines...

src/validation.rs
```diff
@@ -2186,6 +2194,85 @@ fn verify_distribution_behavior(dist_path: &Path) -> Result<Vec<String>> {
...
+                    .unchecked()
+                    .stdout_capture()
+                    .run()
+                    .context(format!("Failed to run `{ldd} {shared_lib}`"))?;
+                let stdout = String::from_utf8_lossy(&output.stdout);
+                // Format of ldd output, for both glibc and musl:
+                // - Everything starts with a tab.
+                // - Most things are "libxyz.so.1 => /usr/lib/libxyz.so.1 (0xabcde000)".
+                // - The ELF interpreter is displayed as just "/lib/ld.so (0xabcde000)".
+                // - glibc, but not musl, shows the vDSO as "linux-vdso.so.1 (0xfffff000)".
```
polarathene: This is not true regarding glibc vs musl for vDSO, just look at the output I showed from Ubuntu and Fedora containers with `ldd`, that's glibc and still includes `linux-vdso.so.1` (these are transitive from glibc).

Realistically, you should only be interested in direct dependencies? As per `patchelf --print-needed` output (_which doesn't add tabbed indents_). That will show what's actually encoded as `DT_NEEDED` entries.

Anything transitive is resolved at runtime, thus `ldd` can be useful for troubleshooting that dependency chain at a runtime environment and filtering for `not found` if anything failed to resolve. Likewise, you could validate the libraries were resolved to the paths you expect.

src/validation.rs
```diff
@@ -2186,6 +2194,85 @@ fn verify_distribution_behavior(dist_path: &Path) -> Result<Vec<String>> {
...
+                // - If a library is listed in DT_NEEDED with an absolute path, or (currently only
+                //   supported on glibc) with an $ORIGIN-relative path, it displays as just
+                //   "/path/to/libxyz.so (0xabcde000)".
+                // - On glibc, if a library cannot be found ldd returns zero and shows "=> not
+                //   found" as the resolution (even if it wouldn't use the => form if found).
+                // - On musl, if a library cannot be found, ldd returns nonzero and shows "Error
+                //   loading shared library ...:" on stderr.
```
polarathene: In glibc you'll often find `ldd` is a bash script (varies by distro) around the actual runtime dynamic linker, which you could call with `--list` to resolve libraries:

```console
# `/usr/bin/ldd` is a bash script wrapper around this:
$ /lib64/ld-linux-x86-64.so.2 --list /lib64/libc.so.6
        /lib64/ld-linux-x86-64.so.2 (0x00007f8ead016000)
        linux-vdso.so.1 (0x00007fff355f6000)

# Failure example:
$ /lib64/ld-linux-x86-64.so.2 --list /root/.local/share/uv/python/cpython-3.13.6-linux-x86_64-gnu/lib/python3.13/lib-dynload/_tkinter.cpython-313-x86_64-linux-gnu.so

/root/.local/share/uv/python/cpython-3.13.6-linux-x86_64-gnu/lib/python3.13/lib-dynload/_tkinter.cpython-313-x86_64-linux-gnu.so: error while loading shared libraries: libtcl8.6.so: cannot open shared object file: No such file or directory
```

As you can see the error message differs from the `not found` that the `ldd` script command outputs.

Alpine has a very simple one for musl:

```console
$ docker run --rm -it alpine
$ cat /usr/bin/ldd
#!/bin/sh
exec /lib/ld-musl-x86_64.so.1 --list "$@"
```

For testing that validates direct deps of a library/executable are found, you should be able to use `patchelf --print-needed` to get a list of deps, then check those against `patchelf --print-rpath` paths, followed by the systems standard search paths (_a match from `ldconfig --print-cache` output should work? Just don't use that from Alpine_). That should be mostly glibc/musl agnostic then? (_With exceptions like `libc.so => /lib/ld-musl-x86_64.so.1`_)

johnthagen:
@geofft Any chance this can be taken over the end? We're currently unable to use `python-build-standalone` due to incompatibilities with PyInstaller + Tk.

Happy to test beta releases of `python-build-standalone` if that would help.

jjhelmus:
> We're currently unable to use `python-build-standalone` due to incompatibilities with PyInstaller + Tk.
Is an ORIGIN-relative `DT_NEEDED` in _tkinter required to resolve the incompatibilities or would adding `DT_RPATH` be sufficient?

johnthagen:
> Is an ORIGIN-relative DT_NEEDED in _tkinter required to resolve the incompatibilities or would adding DT_RPATH be sufficient?

If that was a question directed at me, unfortunately, I'm not sure as I'm only an end user that filed the issue. Perhaps @bwoodsend or someone from the PyInstaller team could answer.

Here are some related discussions:

- #746
- https://github.com/pyinstaller/pyinstaller/issues/9204

bwoodsend:
Sorry, this is quite a long context switch so I don't really remember. I think our criteria is just that `ldd $prefix/lib/python3.9/lib-dynload/_tkinter.cpython-39-x86_64-linux-gnu.so` should resolve to the bundled `libtk`/`libtcl` libraries without relying on inheriting lookup paths from `bin/python` or `libpython.so`. @rokm might know better. Really the best check is just to do a build and we'll see if it breaks.

rokm:
> I think our criteria is just that `ldd $prefix/lib/python3.9/lib-dynload/_tkinter.cpython-39-x86_64-linux-gnu.so` should resolve to the bundled `libtk`/`libtcl` libraries without relying on inheriting lookup paths from `bin/python` or `libpython.so`.

That's indeed the case, and I think either ORIGIN-relative `DT_RPATH` or ORIGIN-relative `DT_NEEDED` would do the trick. Preferably, this would be set on *all* extensions and shared libs provided by `python-build-standalone`, to eliminate the possibility of system-provided libs from slipping in.

FWIW, Anaconda uses `DT_RPATH`, so that should definitely work with PyInstaller. The `DT_NEEDED` came from motivation outlined in https://github.com/pyinstaller/pyinstaller/issues/9204#issuecomment-3171389220, and initially turned out to be problematic for PyInstaller, because it strictly requires the library to be found at the relative path specified by `DT_NEEDED` and we did not preserve that (whereas `DT_RPATH` would, in the absence of library at specified location, fall back to path in `LD_LIBRARY_PATH` that PyInstaller sets). We've fixed that in https://github.com/pyinstaller/pyinstaller/pull/9212, though.

jjhelmus:
> That's indeed the case, and I think either ORIGIN-relative `DT_RPATH` or ORIGIN-relative `DT_NEEDED` would do the trick. Preferably, this would be set on _all_ extensions and shared libs provided by `python-build-standalone`, to eliminate the possibility of system-provided libs from slipping in.

I'm traveling a bit this week and next but will put together a PR to add a `DT_RATH` entry to the `_tkinter` module when I get back.

polarathene:
Hey so I provided review feedback a while back, and at that time I did revise [this comment](https://github.com/nagisa/rust_libloading/issues/161#issuecomment-2994916168) that I was going to update my review feedback to direct to as a reference.

I didn't quite complete that revision to the extent that I wanted to due to a system crash and some data loss. I had a more complete reproduction to show differences in behaviour between glibc and musl systems, but the linked ref seems to cover that (_unless the reproduction revealed some correction or additional caveats I found, I can't quite recall_).

I do recall that behaviour differed between the executable interpreter resolving dependencies vs runtime `dlopen()` behaviour.

`DT_RUNPATH` should be preferred over `DT_RPATH`. For glibc you'll also want to set `DT_RUNPATH` on any vendored library deps you distribute, but on musl `DT_RUNPATH` is effectively treated as `LD_LIBRARY_PATH`/`DT_RUNPATH` where this is not needed (_they inherit the search path override and any attempt to override on their own binary is ignored_). So you may want to keep that caveat in mind.

Regarding `DT_NEEDED`, that shouldn't be an issue with just the library name, no need to prefix origin there, so long as you're using an absolute path the search path from `DT_RUNPATH` should work. I think that got a bit more iffy when dealing with `dlopen()` calls though.

johnthagen:
@jjhelmus I saw that PBS was recently updated to Tcl/Tk 9. Wasn't sure if that had any bearing on the `DT_RPATH` work

- https://github.com/astral-sh/python-build-standalone/releases/tag/20260113

jjhelmus:
> @jjhelmus I saw that PBS was recently updated to Tcl/Tk 9. Wasn't sure if that had any bearing on the `DT_RPATH` work
> 
> * https://github.com/astral-sh/python-build-standalone/releases/tag/20260113

No change in plans, adding a `DT_RPATH` is still on my list but I wanted to finish the Tcl/Tk upgrade before.

polarathene:
Please use `DT_RUNPATH` instead of `DT_RPATH` ([ref](https://github.com/rust-lang/rust/issues/30378#issue-122143220))

johnthagen:
@jjhelmus

> No change in plans, adding a DT_RPATH is still on my list but I wanted to finish the Tcl/Tk upgrade before.

Just checking in to see if there was anything the community could do to help test this. The official [`python`](https://hub.docker.com/_/python) container image dropped support for `bullseye`, so targeting an older glibc using that is no longer possible if a Tk/PyInstaller application wants to target a newer Python, such as 3.14. (#746)

jjhelmus:
> Just checking in to see if there was anything the community could do to help test this. 

If anyone from the community wanted to update this PR or create a new one that added RPATH entries to _tkinter I'd be happy to review. I'll try to get to this soon but have other PBS issues to handle as well.

Issue #761, opened 2025-08-16
[3.14] compression.zstd is slow


Rogdham:
Hello, I'm noticing that the `compression.zstd` in CPython 3.14.0rc2 module has slow performance when using the release assets from this repository.

Local tests show between 4 and 7 times slowdown when compared to building CPython from source, depending on the Zstandard level, and operation made (compression/decompression).

_Maybe related: #327_

## Reproducer

First, download `enwik8.zip` ([official source](https://www.mattmahoney.net/dc/enwik8.zip)), uncompress it at `/tmp/enwik8` (should be 100Mo).

Then, run:

```
python -m timeit -s 'from compression import zstd; DATA = open("/tmp/enwik8","rb").read()' 'zstd.compress(DATA)'
```

| Version | Timing |
|-|-|
| Local build | 1 loop, best of 5: 663 msec per loop |
| cpython-3.14.0rc2+20250814-x86_64-unknown-linux-gnu-install_only_stripped | 1 loop, best of 5: 3.21 sec per loop |
| cpython-3.14.0rc2+20250814-x86_64_v2-unknown-linux-gnu-install_only_stripped | 1 loop, best of 5: 3.2 sec per loop |
| cpython-3.14.0rc2+20250814-x86_64_v3-unknown-linux-gnu-install_only_stripped | 1 loop, best of 5: 3.25 sec per loop |
| cpython-3.14.0rc2+20250814-x86_64-unknown-linux-gnu-pgo+lto-full | 1 loop, best of 5: 3.23 sec per loop |
| cpython-3.14.0rc2+20250814-x86_64-unknown-linux-gnu-freethreaded+pgo+lto-full | 1 loop, best of 5: 3.27 sec per loop |

Tests performed under Archlinux.

zanieb:
Thanks for the report!

I ran
```
python -c "from compression import zstd; print(zstd.CompressionParameter.nb_workers.bounds())"
```

and the result is `(0, 0)` which I think means multithreading is disabled? Can you confirm the bounds are different for your from-source build?

zanieb:
This would do it...

https://github.com/facebook/zstd/blob/e128976193546dceb24249206a02ff8f444f7120/lib/README.md?plain=1#L22

> When building with `make`, by default the dynamic library is multithreaded and static library is single-threaded (for compatibility reasons).

Rogdham:
> and the result is (0, 0) which I think means multithreading is disabled?

[...]

PR #762, opened 2025-08-16
Build zstd with multithreading enabled

zanieb:
I verified this on macOS

```
❯ python -c "from compression import zstd; print(zstd.CompressionParameter.nb_workers.bounds())"
(0, 256)
```

See #761 

I'm not sure if this is all of the performance difference, but using threads should make a significant difference. https://github.com/astral-sh/python-build-standalone/issues/761#issuecomment-3193682600 notes that threads are not used by default, so I think there's more going on — but we should enable this at build time regardless. See also #768.

### DIFF
--- a/src/verify_distribution.py
+++ b/src/verify_distribution.py
@@ -172,6 +172,15 @@ class TestPythonInterpreter(unittest.TestCase):
 
         self.assertEqual(sysconfig.get_config_var("Py_GIL_DISABLED"), wanted)
 
+    @unittest.skipIf(
+        sys.version_info[:2] < (3, 14),
+        "zstd is only available in 3.14+",
+    )
+    def test_zstd_multithreaded(self):
+        from compression import zstd
+
+        assert zstd.CompressionParameter.nb_workers.bounds() == (0, 256)
+
     @unittest.skipIf("TCL_LIBRARY" not in os.environ, "TCL_LIBRARY not set")
     @unittest.skipIf("DISPLAY" not in os.environ, "DISPLAY not set")
     def test_tkinter(self):


zanieb:
We can add a test case that verifies this at runtime.

Rogdham:
This does not solve #761 (I'm having the same timings as on the issue), however it is a good thing that should be done anyways.

I can confirm that this PR adds support for multithreading.

For example, without this PR, the following exception would be raised:
```py
>>> from compression import zstd
>>> zstd.compress(b'foo', options={zstd.CompressionParameter.nb_workers: 2})
  File "<python-input-1>", line 1, in <module>
    zstd.compress(b'foo', options={zstd.CompressionParameter.nb_workers: 2})
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/redacted/.local/share/uv/python/cpython-3.14.0rc1-linux-x86_64-gnu/lib/python3.14/compression/zstd/__init__.py", line 148, in compress
    comp = ZstdCompressor(level=level, options=options, zstd_dict=zstd_dict)
ValueError: compression parameter 'nb_workers' received an illegal value 2; the valid range is [0, 0]
```

Now with this PR (using `cpython-3.14.0rc2-x86_64_v3-unknown-linux-gnu-debug-20250816T1313.tar.zst` from GitHub action build artefacts):
```py
>>> from compression import zstd
>>> zstd.compress(b'foo', options={zstd.CompressionParameter.nb_workers: 2})
b'(\xb5/\xfd \x03\x19\x00\x00foo'
```

PR #763, opened 2025-08-16
Pin all of the actions

zanieb:


woodruffw:
With this we can also remove the zizmor config that was allowing `astral-sh/*` and `actions/*` to be ref-pinned only. I'll send a quick PR for that 🙂

PR #767, opened 2025-08-16
only define MODULE_NAME for sqlite on python <= 3.9

dtrodrigues:
Starting in Python 3.10, -DMODULE_NAME="sqlite3" is no longer needed when building _sqlite3.
https://github.com/python/cpython/issues/87610 / https://github.com/python/cpython/pull/24801


zanieb:
Thanks!

PR #780, opened 2025-08-30
On macOS, add 64 bytes of headerpad to accommodate code signatures

### LINKED ISSUE #749: Unable to code sign _tkinter library on x86 Mac version due to insufficient headerpad
Hi, we're planning to bundle these standalone versions (Windows, x86 Linux, x86 Mac, and M1 Mac) with the next jGRASP release.

From cpython-3.13.6+20250808-x86_64-apple-darwin-install_only I was able to code sign everything necessary except for lib/python3.13/lib-dynload/_tkinter.cpython-313-darwin.so

I get the error "can't allocate code signature data ... because larger updated load commands do not fit (the program must be relinked using a larger -headerpad value).

The command used is: codesign -f -o runtime -s "cert_name" file_name

The corresponding Mac aarch64 library does not have this problem.

We develop on a fairly old macOS version (11.6.1) for ease of compatibility.

geofft:
Closes #749.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -365,6 +365,16 @@ if [[ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_12}" && "${TARGET_TRIPLE}" = "ppc64le
     LDFLAGS="${LDFLAGS} -Wl,--no-tls-get-addr-optimize"
 fi
 
+# We're calling install_name_tool -add_rpath on extension modules, which
+# eats up 0x20 bytes of space in the Mach-O header, and we need to make
+# sure there's still enough room to add a code signature (0x10 bytes) on
+# non-arm64 where there's no automatic ad-hoc signature. We are somehow
+# on a toolchain that doesn't make sure there's enough space by default
+# so give it plenty of space.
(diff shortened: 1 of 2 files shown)


geofft:
[Failing tests from previous run](https://github.com/astral-sh/python-build-standalone/actions/runs/17361700681?pr=780), as expected on x86_64 only:

```
Error: errors found
  error: python/install/lib/python3.13/lib-dynload/_ctypes_test.cpython-313d-darwin.so: Insufficient headerpad between end of load commands 0x648 and beginning of code 0x650
  error: python/install/lib/python3.13/lib-dynload/_tkinter.cpython-313d-darwin.so: Insufficient headerpad between end of load commands 0x658 and beginning of code 0x660
```

indygreg (approved):
This looks good.

What's also surprising about this is that usually segments are aligned on 4k or 16k boundaries. i.e. there's usually a significant gap between load commands and the next data.

And load commands are usually nowhere close to 4k in size.

Again, I'm surprised there isn't a sufficient gap for the code signature load command.

But surprised or not, having a static check to ensure existence of a gap keeps us honest and is necessary.

PR #781, opened 2025-08-31
Run `cargo fmt` and `cargo clippy --fix`

geofft:
This is from cargo 1.88, since 1.89 pulls in let-chains, and whatever version of Rust is in GitHub CI doesn't support them last time I tried. (We should upgrade it at some point, though, since our Rust usage is all internal.)

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -13,7 +13,7 @@ use std::{
 use url::Url;
 use {
     crate::json::parse_python_json,
-    anyhow::{anyhow, Result},
+    anyhow::{Result, anyhow},
     once_cell::sync::Lazy,
     pep440_rs::VersionSpecifier,
     std::{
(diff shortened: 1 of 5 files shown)


src/release.rs
```diff
@@ -628,7 +628,7 @@ static LLVM_URL: Lazy<Url> = Lazy::new(|| {
 /// Returns the path to the top-level `llvm` directory.
 pub async fn bootstrap_llvm() -> Result<PathBuf> {
     let url = &*LLVM_URL;
-    let filename = url.path_segments().unwrap().last().unwrap();
+    let filename = url.path_segments().unwrap().next_back().unwrap();
```
geofft, line 631: "warning: called `Iterator::last` on a `DoubleEndedIterator`; this will needlessly iterate the entire iterator"

https://rust-lang.github.io/rust-clippy/master/index.html#double_ended_iterator_last

PR #797, opened 2025-09-17
Upgrade OpenSSL to 3.5.3

geofft:


geofft:
cc @Edward-Knight

Edward-Knight (approved):
LGTM!
I see the vlenb patch hit in 3.5.3 as well as a fix for the `OPENSSL_VERSION_NUMBER` that now (correctly) makes it end in `0xf` (for **f**inal I presume?)

geofft:
> as well as a fix for the OPENSSL_VERSION_NUMBER that now (correctly) makes it end in 0xf (for final I presume?)

Turns out this was a mistake that was only present in 3.5.3. The 1.x series was [documented](https://docs.openssl.org/1.1.1/man3/OPENSSL_VERSION_NUMBER/#synopsis) as: "The status nibble has one of the values 0 for development, 1 to e for betas 1 to 14, and f for release." But the 3.x series was always [documented](https://docs.openssl.org/3.5/man3/OpenSSL_version/#macros) as: "OPENSSL_VERSION_NUMBER is a combination of the major, minor and patch version into a single integer 0xMNN00PP0L". 3.5.3 was the only version that used 0xF; they reverted that for 3.5.4 and this didn't get into any other releases. (See openssl/project#1621 and linked issues.)

Edward-Knight:
How 0xfun

PR #694, opened 2025-07-10
Enable FTS5 in SQLite

zanieb:

Issue #309, opened 2024-08-29
sqlite SQLITE_ENABLE_DBSTAT_VTAB option 


cjrh:
The sqlite option SQLITE_ENABLE_DBSTAT_VTAB enables the `dbstat` virtual table that can show table sizes. The Python on my distro (Mint) was linked to a sqlite that had been compiled with this option, but it seems the default `build-linux.py` configuration does not set it.

I was going to make a PR to add it to `cpython-unix/extension-modules.yml` like so:

```yml
_sqlite3:
  sources:
    - _sqlite/connection.c
    - _sqlite/cursor.c
    - _sqlite/microprotocols.c
    - _sqlite/module.c
    - _sqlite/prepare_protocol.c
    - _sqlite/row.c
    - _sqlite/statement.c
    - _sqlite/util.c
  sources-conditional:
    - source: _sqlite/blob.c
      minimum-python-version: "3.11"
    - source: _sqlite/cache.c
      maximum-python-version: "3.10"
  includes-deps:
    - include
  includes:
    - Modules/_sqlite
  defines:
    - "MODULE_NAME=\\\"sqlite3\\\""
  defines-conditional:
    # Require dynamic binaries to load extensions. Cannot load on iOS.
    # 3.11+ uses opt in. <3.11 uses opt out.
    - define: PY_SQLITE_ENABLE_LOAD_EXTENSION=1
      targets:
        - .*-apple-darwin
        # TODO this should likely be restricted to gnu since musl is statically
        # linked. But this would break verification code. So enabled for
        # backwards compatibility.
        - .*-unknown-linux-.*
    # This enables the sqlite3_dbstat virtual table, which is typically
    # enabled on linux distro builds.
    # https://www.sqlite.org/dbstat.html
    - define: SQLITE_ENABLE_DBSTAT_VTAB=1                                # <------- HERE
      targets:
        - .*-apple-darwin
        - .*-unknown-linux-.*
    - define: SQLITE_OMIT_LOAD_EXTENSION=1
      targets:
        - .*-ios
  links:
    - sqlite3
```

However, this doesn't work. In the resulting python build the compile option has not been set:

```
$ ./build-linux.py --python cpython-3.12
<snip compiler output>
$ cd dist
$ tar --zstd -xf cpython-3.12.5-x86_64-unknown-linux-gnu-noopt-20240824T1350.tar.zst 
$ ./python/install/bin/python
[...]

PR #800, opened 2025-09-18
windows: Sync SQLite build flags from the UNIX build

geofft:
Notably this adds
* SQLITE_ENABLE_FTS3_PARENTHESIS (syntax change, see #550)
* SQLITE_ENABLE_DBSTAT_VTAB (#309)
* SQLITE_ENABLE_GEOPOLY (historically present on UNIX, maybe see #694)

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -568,6 +568,36 @@ def hack_project_files(
         rb"<SqlitePatchVersion>%s</SqlitePatchVersion>" % sqlite3_version_parts[3],
     )
 
+    # Please try keep these in sync with cpython-unix/build-sqlite.sh
+    sqlite_build_flags = {
+        b"SQLITE_ENABLE_DBSTAT_VTAB",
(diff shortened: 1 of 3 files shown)


cpython-windows/build.py
```diff
@@ -568,6 +568,36 @@ def hack_project_files(
...
 
+    # Please try keep these in sync with cpython-unix/build-sqlite.sh
+    sqlite_build_flags = {
+        b"SQLITE_ENABLE_DBSTAT_VTAB",
+        b"SQLITE_ENABLE_FTS3",
+        b"SQLITE_ENABLE_FTS3_PARENTHESIS",
+        b"SQLITE_ENABLE_FTS4",
+        b"SQLITE_ENABLE_FTS5",
+        b"SQLITE_ENABLE_GEOPOLY",
+        b"SQLITE_ENABLE_RTREE",
```
zanieb, line 579: It sounds like this list should live somewhere else!

PR #802, opened 2025-09-18
CI: Disable Docker build summaries and artifacts

geofft:
The interesting work for our build happens outside of the Docker build, so these just make the GitHub Actions summary page noisier.

https://github.com/docker/build-push-action#environment-variables


geofft:
Failing because of #803 but I think this does what I want in terms of the summary output.

PR #804, opened 2025-09-18
Fix AsyncioDebug section detection on Python 3.14

### LINKED ISSUE #717: Error retrieving tasks: AsyncioDebug section not available, in cpython-3.14.0rc1-macos-aarch64-none tag/20250723
Hello,

I'm following the example from https://docs.python.org/3.14/whatsnew/3.14.html#asyncio-introspection-capabilities and it fails with

    Error retrieving tasks: AsyncioDebug section not available

Please check this workflow run on matrix of macos, ubuntu, windows: https://github.com/aleksandr-vin/Python_asynico_introspection_test_MacOS/actions/runs/16673137842/job/47193715249 -- you will see that `python -m asyncio pstree` is producing this error on macos with python 3.14.0rc1 installed via uv and works fine on other platforms or when python is installed via actions/setup-python.

geofft:
Fixes #717.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -319,6 +319,13 @@ if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_12}" ]; then
     patch -p1 -i ${ROOT}/patch-test-embed-prevent-segfault.patch
 fi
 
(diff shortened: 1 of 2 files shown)


zanieb (approved):

geofft:
Manually tested and it seems to work! I should write some tests (or figure out if upstream's test suite is enough).

PR #807, opened 2025-10-01
Bump OpenSSL 3.5.3 -> 3.5.4

geofft:


geofft:
https://openssl-library.org/post/2025-09-30-release-announcement/

> Changes and CVEs fixed in 3.5.4:
>
> * [CVE-2025-9230](https://www.openssl.org/news/vulnerabilities.html#CVE-2025-9230) - Fix Out-of-bounds read & write in RFC 3211 KEK Unwrap.
> * [CVE-2025-9231](https://www.openssl.org/news/vulnerabilities.html#CVE-2025-9231) - Fix Timing side-channel in SM2 algorithm on 64-bit ARM.
> * [CVE-2025-9232](https://www.openssl.org/news/vulnerabilities.html#CVE-2025-9232) - Fix Out-of-bounds read in HTTP client no_proxy handling.
> * Reverted the synthesised `OPENSSL_VERSION_NUMBER` change for the release builds, as it broke some exiting applications that relied on the previous 3.x semantics, as documented in `OpenSSL_version(3)`.

None of these CVEs look particularly concerning. The first two are rated "moderate" but the first one is for CMS (S/MIME etc.) which I think isn't even exposed to Python, and the second one is for a configuration that isn't usable by default and you'd have to write custom compiled code to enable. The third one is rated "low". So, no rush to cut a release for these IMO.

PR #817, opened 2025-10-09
Bump stable Python versions: 3.12.12, 3.11.14, 3.10.19 and 3.9.24

edgarrmondragon:


edgarrmondragon:
Ah, I forgot to update the sizes. Will push them shortly unless you're already working on that @zanieb 🙂

geofft:
This is in the 20251010 release which will be in uv soon (astral-sh/uv#16229), thanks again!

PR #819, opened 2025-10-14
Add support for Python 3.15

zanieb:


zanieb:
I encountered a test failure during PGO, tracked in https://github.com/python/cpython/issues/140125 — ignoring the test for now.

PR #831, opened 2025-10-27
Use latest LLVM for musl builds (fixes #829)

### LINKED ISSUE #829: Upgrade musl builds to latest LLVM
Right now, we're using LLVM for musl 

https://github.com/astral-sh/python-build-standalone/blob/01aa80641edfca8efa1a8f7c3b9bb0405bde1512/pythonbuild/utils.py#L480-L482

There are various problems that aren't worth investigating until we're on the newest compiler version. We should upgrade to LLVM 20.

geofft:


jjhelmus:
LLVM for musl was upgraded in #916. Is this PR still needed?

PR #833, opened 2025-10-28
include existing LDFLAGS on macos when building tk

### LINKED ISSUE #737: build-tk.sh drops EXTRA_TARGET_LDFLAGS for macos
```bash
CFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC"
LDFLAGS="${EXTRA_TARGET_LDFLAGS}"

if [[ "${PYBUILD_PLATFORM}" = macos* ]]; then
    CFLAGS="${CFLAGS} -I${TOOLS_PATH}/deps/include -Wno-availability"
    CFLAGS="${CFLAGS} -Wno-deprecated-declarations -Wno-unknown-attributes -Wno-typedef-redefinition"
    LDFLAGS="-L${TOOLS_PATH}/deps/lib"
    EXTRA_CONFIGURE_FLAGS="--enable-aqua=yes --without-x"
else
    EXTRA_CONFIGURE_FLAGS="--x-includes=${TOOLS_PATH}/deps/include --x-libraries=${TOOLS_PATH}/deps/lib"
fi

CFLAGS="${CFLAGS}" CPPFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ./configure \
```

I think this is a bug and it should be `LDFLAGS="${LDFLAGS} -L${TOOLS_PATH}/deps/lib"`, but I don't want to roll this into #676 just in case it changes behavior in an interesting way.

jjhelmus:
closes #737

### DIFF
--- a/cpython-unix/build-tk.sh
+++ b/cpython-unix/build-tk.sh
@@ -26,7 +26,7 @@ LDFLAGS="${EXTRA_TARGET_LDFLAGS}"
 if [[ "${PYBUILD_PLATFORM}" = macos* ]]; then
     CFLAGS="${CFLAGS} -I${TOOLS_PATH}/deps/include -Wno-availability"
     CFLAGS="${CFLAGS} -Wno-deprecated-declarations -Wno-unknown-attributes -Wno-typedef-redefinition"
-    LDFLAGS="-L${TOOLS_PATH}/deps/lib"
+    LDFLAGS="${LDFLAGS} -L${TOOLS_PATH}/deps/lib"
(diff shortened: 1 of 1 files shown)


jjhelmus:
This does not change the resulting artifact so I'm not seeing anything worthwhile to test. 

The existing behavior was excluding the flags passed via `EXTRA_TARGET_LDFLAGS` when linking `libtk` but these flags are also included in CFLAGS which are correctly passed. 

I checked the logs of this build against those prior to this change and with the changes the flags in questions (e.g. `-arch arm64 -mmacosx-version-min=11.0 -isysroot ...`) are duplicated with this change.

PR #834, opened 2025-10-29
Exercise tests_curses_interactive in CI

### LINKED ISSUE #697: Run test_curses_interactive in CI

jjhelmus:
Use a barebones TERM setting to allow test_curses_interactive to run in CI.

closes #697

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -417,3 +417,4 @@ jobs:
         env:
           MATRIX_RUN: ${{ matrix.run }}
           MATRIX_LIBC: ${{ matrix.libc }}
+          TERM: xterm


zanieb:
You should probably run this on a subset of the matrix, e.g., `python:3.13`, otherwise we're just wasting a lot of compute.

Issue #841, opened 2025-11-01
Remove Python 3.9


geofft:
Now that Python 3.9 is EOL, we should remove it from the build. (I mentioned this in the [20251031 release notes](https://github.com/astral-sh/python-build-standalone/releases/tag/20251031).)

hugovk:
Please see https://github.com/astral-sh/python-build-standalone/pull/845 for a first pass.

jjhelmus:
#845 removed Python 3.9 from the build matrix and updated the build tooling so that running it now required Python 3.10+.  Closing.

PR #845, opened 2025-11-05
Drop support for EOL Python 3.9

hugovk:
First pass for https://github.com/astral-sh/python-build-standalone/issues/841.

There's some `3.9` still left in https://github.com/hugovk/python-build-standalone/blob/main/cpython-unix/targets.yml because seven targets _only_ support 3.9, like `aarch64-apple-ios` and `arm64-apple-tvos`:

https://github.com/hugovk/python-build-standalone/blob/2181bddcc918f14275232ad6d1365035f2ae75c4/cpython-unix/targets.yml#L114-L119

Should they be removed? Or upgraded?

Also ran `ruff check --select UP --target-version py310 --fix` and `run check --fix`. There's a lot of `--unsafe-fixes` that can be applied, but they can be in their own PR to avoid making this too big.


jjhelmus:
This looks great. Thanks @hugovk 

> Should they be removed? Or upgraded?

I would suggest removing these targets. They have not been built in CI for a long while and AFAIK have never been released.

hugovk:
I had removed iOS/tvOS/watchOS stuff from patch files such as https://github.com/astral-sh/python-build-standalone/blob/bdabbfb560c3b42341e00283957bae4dcdc5c36f/cpython-unix/patch-apple-cross.patch#L9-L17 but they didn't apply so need doing with more care. So I left it out from this already big PR, they can be dealt with another time.

jjhelmus (approved):

PR #851, opened 2025-11-11
Define additional build flags in targets.yaml, including those split between debug/standard

jjhelmus:
This is a work in progress. I've tried a few different approaches to adding in these flags. Want to test this on a different system before filling in the details and expanding the flags to other targets.

### DIFF
--- a/cpython-unix/build.py
+++ b/cpython-unix/build.py
@@ -64,7 +64,7 @@ MACOS_ALLOW_SYSTEM_LIBRARIES = {"dl", "m", "pthread"}
 MACOS_ALLOW_FRAMEWORKS = {"CoreFoundation"}
 
 
-def add_target_env(env, build_platform, target_triple, build_env):
+def add_target_env(env, build_platform, target_triple, build_env, build_options):
     add_env_common(env)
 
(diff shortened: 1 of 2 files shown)


zanieb:
Don't we want to share these across all targets? Should we be defining them in target-specific configuration?

jjhelmus:
The optimization flags should apply across all platforms but the hardening flags will be platform specific. 

I tried a few different approaches but was not entirely happy with any of them. The approaches I tried:

* Have `build-*` source a setup script that sets buildflags
This is nice as the script can adjust flags based upon OS and architecture as well as set up some common variables (`ROOT`, `PATH`). 

The downside is that the build options are passed to the build script so adjust the flags for debug vs LTO, etc is not possible. Additionally each script needs to source this file which requires changing, and remember to change a number of scripts.

* Introduce static build flags in a new yaml file
A new file can be used to introduce a set of build flags across all platforms. 

Downsides are this this introduced a second location where static flags are introduced and does not offer options for adjusting these flags for OS or architecture.

jjhelmus:
Closing this in favor of #852 which only address the optimization flag. 

I would like to revisit how build flags are set during the build process in more detail in the future.

PR #856, opened 2025-11-14
Do not strip VC redists

### LINKED ISSUE #855: `llvm-strip` invalidates digital signature of MSVC redists
When creating the stripped release, `llvm-strip` is called over ALL dlls, including those that are not from the python build. `vcruntime140.dll` and `vcruntime140_1.dll` are provided by Microsoft in order to ensure that the compiled program can run on a different machine. The stripping should be limited to removing the python debug symbols, and leaving the redist files alone.

The signature can be checked in powershell with:
`Get-AuthenticodeSignature <PATH>`
Or in the file explorer by right clicking the file --> clicking properties --> clicking digital signatures

See that the signature is intact in the non-stripped variant: `https://github.com/astral-sh/python-build-standalone/releases/download/20240814/cpython-3.11.9%2B20240814-x86_64-pc-windows-msvc-install_only.tar.gz`
And the signature is removed in the stripped variant: `https://github.com/astral-sh/python-build-standalone/releases/download/20240814/cpython-3.11.9%2B20240814-x86_64-pc-windows-msvc-install_only_stripped.tar.gz`

For each VC dll, stripping only removes 11 KB from the file. So skipping the strip on this file will barely have an impact on the overall distribution size.

Keeping the signature intact will fix certain cases where the python build is not redistributable

neilmehta24:
Closes #855 

Please see the above issue description for why this addition is proposed

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -511,8 +511,12 @@ pub fn convert_to_stripped<W: Write>(
                 | FileKind::Pe32
                 | FileKind::Pe64)
         ) {
-            data = llvm_strip(&data, llvm_dir)
-                .with_context(|| format!("failed to strip {}", path.display()))?;
+            // Skip stripping MSVC runtime DLLs
+            let filename = path.file_name().and_then(|n| n.to_str());
+            if !matches!(filename, Some("vcruntime140.dll" | "vcruntime140_1.dll")) {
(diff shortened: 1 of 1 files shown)


jjhelmus:
I agree that symbols should not be stripped from these DLLs. Thanks for the fix @neilmehta24.

PR #857, opened 2025-11-14
Make the Unix build truly relocatable

mrmathematica:
Currently, the python build on Unix is not really relocatalbe. When installing, uv further patches the `_sysconfigdata_` file in uv/crates/uv-python/src/sysconfig/mod.rs `update_sysconfig` and `patch_sysconfigdata` function. Since `_sysconfigdata_` file is a python file, we can let it dynamically decide the install path here in python-build-standalone, instead of rely on further patch down the line.

I see from uv/crates/uv-python/src/sysconfig/mod.rs, it also "remove any references to `-isysroot` in a whitespace-separated string". If needed I can add that part of logic to here as well.

The aim for this change is that, for python install to be able to be copied into any place after install, it should still work. Doing so here instead of inside uv should make it more avialable.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -843,6 +843,7 @@ fi
 cat > ${ROOT}/hack_sysconfig.py << EOF
 import json
 import os
+import re
 import sys
 import sysconfig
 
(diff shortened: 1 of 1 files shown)


jjhelmus:
@mrmathematica Thanks for putting this together. 

Working towards a more relocatable distribution is a goal of this project but we also want to balance this with not deviating too far from upstream. Rather than modify sysconfigdata as part of this projects build script a preferred approach would be apply a patch to CPython itself to performs these modification as part of the standard build process. If these modification are part of building CPython directly the patch can be submitted upstream for inclusion in a future release of CPython. See #861 for an example of this type of work where portions of the RPATH adjusts in the build script are migrated to a patch.

If this is something you are interested in working on, I'd be happy to assist.

mrmathematica:
@jjhelmus  
Do you mean a patch file here first, or a patch to CPython itself? A patch to CPython, even if merged, will likely only be adopted for python 3.15+, thus not helping all existing python versions. If you mean a patch file applied in this repo, I think we can work together for that.

jjhelmus:
A patch here first is what I was referring to.

mrmathematica:
OK @jjhelmus  I give that a [try](https://github.com/mrmathematica/python-build-standalone/commit/e9388a57a9ebb5e5db5d99c7432654a18b0a3c59).

I see there can be all kinds of downstream places that touches `_sysconfigdata_` again, including https://github.com/python/cpython/blob/a03efb533a58fd13fb0cc7f4a5c02c8406a407bd/Mac/BuildScript/build-installer.py#L1360C1-L1385C15 as mentioned in one of the touched file. All such places need to be modified to work with the new `_sysconfigdata_.py` file format and f-string inside dict, which is kind of fragile.

On the otherhand, the change in this PR is much more robust. It happened at the same place in this repo where `_sysconfigdata_.py` is hacked, which is the last step in the process.

mrmathematica:
Hi @jjhelmus  I updated https://github.com/astral-sh/python-build-standalone/pull/923 . Can you run the tests again? With code which approach do you think is more elegent?

PR #860, opened 2025-11-19
Update to CPython 3.15.0a2

jjhelmus:
Update the cpython-3.15 target from 3.15.0a1 -> 3.15.0a2


zanieb:
> Exception: missing extension modules from YAML: _math_integer

jjhelmus:
Dropping these links here for my own benefit:

* [What's new in Python 3.15](https://docs.python.org/3.15/whatsnew/3.15.html)
* [Python 3.15 changelog](https://docs.python.org/3.15/whatsnew/changelog.html)

`_math_interger` is a new stdlib extension module introduced with the addition of `math.integer`. See [PEP 791](https://peps.python.org/pep-0791/) and python/cpython#133909

PR #861, opened 2025-11-21
POC: Add a relocatable build flag

zanieb:
The goal here is to construct two upstreamable patches

1. Add a `--enable-relocatable` configure flag to gate new behavior
2. Add linker options for relocatablity (i.e., rpath) to the CPython build behind the above flag

As a PoC, we apply these patches here for Python 3.15 and skip our post-build relocatability patchelf invocations.

### DIFF
--- a/cpython-unix/patch-relocatable-link.patch
+++ b/cpython-unix/patch-relocatable-link.patch
@@ -1,9 +1,23 @@
+commit 34915211b42a64082e0470ebb9a8c4f2ebb20512
+Author: Zanie Blue <contact@zanie.dev>
+Date:   Fri Nov 21 15:44:42 2025 -0600
+
+    Set linker arguments when `--enabled-relocatable` is used
+
 diff --git a/Makefile.pre.in b/Makefile.pre.in
-index 6c9cba59e27..39fc2e911d0 100644
+index 6c9cba59e27..6c89690c35e 100644
 --- a/Makefile.pre.in
 +++ b/Makefile.pre.in
-@@ -1028,7 +1028,12 @@ libpython3.so:	libpython$(LDVERSION).so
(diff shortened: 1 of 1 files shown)


cpython-unix/patch-relocatable-link.patch
```diff
@@ -0,0 +1,70 @@
...
+@@ -13749,7 +13749,13 @@ then
+ 	    LINKFORSHARED="-Wl,-E -Wl,+s";;
+ #	    LINKFORSHARED="-Wl,-E -Wl,+s -Wl,+b\$(BINLIBDEST)/lib-dynload";;
+ 	Linux-android*) LINKFORSHARED="-pie -Xlinker -export-dynamic";;
+-	Linux*|GNU*) LINKFORSHARED="-Xlinker -export-dynamic";;
++	Linux*|GNU*)
++		LINKFORSHARED="-Xlinker -export-dynamic"
++		# Add rpath for relocatable builds
++		if test "x$enable_relocatable" = xyes; then
++			LINKFORSHARED="$LINKFORSHARED -Wl,-rpath=\$ORIGIN/../lib"
```
jjhelmus:
```suggestion
+			LINKFORSHARED="$LINKFORSHARED -Wl,-rpath,'\$\$ORIGIN/../lib'"
```

jjhelmus:
A question that I suspect will get asked when upstreaming this, is what should be done to accommodate third-party libraries such as libtcl, libtk, libz in relocatable builds.

With the exception of libtcl/tk these are statically linked but this patch should support cases where these are dynamically linked. Adding a RPATH entry to the extensions with a relative path to `lib` (`$ORIGIN/../..`) would be good but not required as bin/python includes an RPATH. 

I could also see an opportunity to hash the SO name of third-party libraries on Linux to protect against clashes with system libraries. This approach is used by `auditwheel` when bundling libraries into the wheel. macOS has two-level namespaces which makes this largely unnecessary.

PR #890, opened 2025-11-25
enable building aarch64-linux targets from macOS

jjhelmus:
Add the ability to build for aarch64-unknown-linux-{musl,gnu} platforms on an arm64 macOS host using Docker.

For example using:
    ./build-linux.py --target-triple aarch64-unknown-linux-musl --python cpython-3.14

### DIFF
--- a/cpython-unix/build-main.py
+++ b/cpython-unix/build-main.py
@@ -79,7 +79,7 @@ def main():
         "--no-docker",
         action="store_true",
         default=True if sys.platform == "darwin" else False,
-        help="Disable building in Docker",
+        help="Disable building in Docker on Linux hosts.",
     )
     parser.add_argument(
         "--serial",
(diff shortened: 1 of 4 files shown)


cpython-unix/build-main.py
```diff
@@ -129,9 +129,26 @@ def main():
 
     musl = "musl" in target_triple
 
+    # Linux targets can be built on a macOS host using Docker.
+    building_linux_from_macos = sys.platform == "darwin" and "linux" in target_triple
+    if building_linux_from_macos:
+        print("Note: Using Docker to build for Linux on macOS")
+        args.no_docker = False
```
zanieb: This feels a little dubious as an implementation. I'd expect us not to mutate `args` and I'd expect `--no-docker` to cause a failure in this case?

PR #896, opened 2025-12-01
Use the symlink-resolved location of Python in getpath on 3.14+

### LINKED ISSUE #380: Virtualenvs created from links are broken
Unlike a system interpreter or a pyenv installation, pbs can't create venv from a symlink to the pbs installation location. PYTHONHOME is in the wrong location (that of the link, i think), which causes the encoding error.

I've only tested linux, i expect windows isn't affected (different venv mechanism, no symlinks by default), but i expect macos is affected, too.

```dockerfile
FROM ubuntu

RUN apt update && apt install -yy wget curl tar python3

# Install pbs in one dir
RUN mkdir /root/a
WORKDIR /root/a
RUN wget https://github.com/indygreg/python-build-standalone/releases/download/20241016/cpython-3.12.7+20241016-x86_64_v3-unknown-linux-gnu-install_only_stripped.tar.gz
RUN tar xf cpython-3.12.7+20241016-x86_64_v3-unknown-linux-gnu-install_only_stripped.tar.gz

# Go to another dir
RUN mkdir /root/b
WORKDIR /root/b

# Symlink the system interpreter for a level of indirection
RUN ln -s /usr/bin/python3.12 python3.12
RUN /usr/bin/python3.12 -m venv --without-pip python3.12-venv-direct
RUN ./python3.12 -m venv --without-pip python3.12-venv-link

# Symlink the pds interpreter for a level of indirection
RUN ln -s /root/a/python/bin/python3 pbs3.12
RUN /root/a/python/bin/python -m venv --without-pip pbs3.12-venv-direct
RUN ./pbs3.12 -m venv --without-pip pbs3.12-venv-link

CMD ["bash", "-c", "./python3.12-venv-direct/bin/python -c 'import sys; print(1, sys._base_executable)' \
    && ./python3.12-venv-link/bin/python -c 'import sys; print(2, sys._base_executable)' \
    && ./pbs3.12-venv-direct/bin/python -c 'import sys; print(3, sys._base_executable)' \
    && ./pbs3.12-venv-link/bin/python -c 'import sys; print(4, sys._base_executable)'"]
```

```
1 /usr/bin/python3.12
2 /usr/bin/python3.12
3 /root/a/python/bin/python3.12
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = './pbs3.12-venv-link/bin/python'
  isolated = 0
  environment = 1
  user site = 1
  safe_path = 0
  import site = 1
  is in build tree = 0
  stdlib dir = '/install/lib/python3.12'
  sys._base_executable = '/root/a/python/bin/python3.12'
  sys.base_prefix = '/install'
  sys.base_exec_prefix = '/install'
  sys.platlibdir = 'lib'
  sys.executable = '/root/b/pbs3.12-venv-link/bin/python'
  sys.prefix = '/install'
  sys.exec_prefix = '/install'
  sys.path = [
    '/install/lib/python312.zip',
    '/install/lib/python3.12',
    '/install/lib/python3.12/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007dce45bbb740 (most recent call first):
  <no Python frame>
```

### LINKED ISSUE #713: os.execl fails to execute Python
From a python script, attempting to run `os.execl` with Python as the executable fails because Python seems to locate `/install` as its path instead of the real path. This installation was installed using uv, but I'm still experiencing this issue.

In the real world, this affects the [yocto poky project](https://github.com/yoctoproject/poky/blob/bd4625cd4db0f02162092d85aeab3023914f768a/bitbake/lib/bb/server/process.py#L603) which works fine with other Python distributions but fails to run with this one. 

Example:

```python
import sys
import os

os.execl(sys.executable, "test")
```
Output:

```
$ python test.py
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'test'
  isolated = 0
  environment = 1
  user site = 1
  safe_path = 0
  import site = 1
  is in build tree = 0
  stdlib dir = '/install/lib/python3.11'
  sys._base_executable = '/usr/bin/test'
  sys.base_prefix = '/install'
  sys.base_exec_prefix = '/install'
  sys.platlibdir = 'lib'
  sys.executable = '/usr/bin/test'
  sys.prefix = '/install'
  sys.exec_prefix = '/install'
  sys.path = [
    '/install/lib/python311.zip',
    '/install/lib/python3.11',
    '/install/lib/python3.11/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007f997d316740 (most recent call first):
  <no Python frame>
```

Using another python distribution (e.g. from Ubuntu or PyEnv) that has the correct compiled in paths:

```
$ python test.py 
Python 3.11.9 (main, Jan  8 2025, 12:47:30) [GCC 9.4.0] on linux
>>> 
```

geofft:
Mostly fixes #380, and also fixes #713 because we prefer the actual
binary path to argv[0].

### DIFF
--- a/src/verify_distribution.py
+++ b/src/verify_distribution.py
@@ -4,12 +4,12 @@
 
 import importlib.machinery
 import os
-from pathlib import Path
 import struct
 import subprocess
 import sys
 import tempfile
 import unittest
+from pathlib import Path
 
 TERMINFO_DIRS = [
     "/etc/terminfo",


jjhelmus:
Passing the path the the executable via the `library` variable in `getpath.py` has a smell as the variable is described as being the path to the dylib/DLL/so. When static linking there is no shared library so I see the logic in passing the path to the binary (as it is a DSO) itself.

jjhelmus:
With this patch the directory pointed to by `home` in the pyvenv.cfg is still searched if the library search fails (but it should not fail). 

The change from https://github.com/python/cpython/issues/106045#issuecomment-2594628161 (below) would update this path to the directory of the base_executable when it exists. I'm somewhat partial to that solution in addition to this change.

``` diff
@@ -411,6 +412,9 @@ def search_up(prefix, *landmarks, test=isfile):
                             if isfile(candidate):
                                 base_executable = candidate
                                 break
+                if base_executable and isfile(base_executable):
+                    # Update the executable directory to be based on the resolved base executable
+                    executable_dir = real_executable_dir = dirname(base_executable)
             # home key found; stop iterating over lines
             break
 ```

**Comment by jjhelmus** [comment]
I've found following helpful check/reproduce/debug this issue:
```
./configure --prefix=/example
make -j
make DESTDIR=<some_real_path> install
cd <some_real_path>
./example/bin/python3.14 # can be used to create symlinks/venv/etc 
```

zanieb:
Did you mean to close this?

jjhelmus:
> Did you mean to close this?

Nope, oops

geofft:
I think adding both this patch and the one in #903 is a good idea, though even with both patches I would consider it not quite a complete fix.

I've added some test cases based on #380 and #713. The patch in #903 by itself passes the tests for #380 but not the one in #713. (On the other hand the patch in #903 works on 3.11+.)

Getting this patch to work for 3.11-3.13 isn't hard, but I'm not sure if I will get to that today.

zanieb:
Includes #903?

jjhelmus:
> Includes #903?

Yes, this supersedes #903

geofft:
Yes, to fix the aarch64 failures.

PR #901, opened 2025-12-03
Increase recursion depth for `test_json` from 200k to 500k

zanieb:
Testing https://github.com/python/cpython/issues/140125#issuecomment-3607044322


zanieb:
Branch PGO took 39.6s
HEAD PGO took 37.5s

The total build time was 2m longer but I don't see why. It might just be runner noise.

zanieb:
(Running again with 3.14 to confirm)

PR #902, opened 2025-12-03
Futz with the macOS CI target definitions

zanieb:


zanieb:
I noticed that the crate build was reporting that it was x86_64 on main but the runner is actually aarch64 (and the crate is built as an aarch64 target)

PR #903, opened 2025-12-04
patch python to allow venv creation from symlinks

jjhelmus:
Patch CPython to allow venvs to be created from symlinks. When inside a venv and base_executable is passed available. Use the resolved path as the executable_dir rather than the one specified by home in pyvenv.cfg


jjhelmus:
superseded by #896

PR #907, opened 2025-12-05
getpath: Work around exciting aarch64 pointers

geofft:


geofft:
Gotta think about this harder....

PR #914, opened 2025-12-08
Correctly set install_name for macOS libpython

keith:
Previously the `install_name` for libpython on macOS was just the prefix
`/install/lib/...` which was knowingly invalid. The post-build scripts
fixed this for `python3` itself, but don't fix the dylib itself for the
case where downstream users of the toolchain try to link libpython. Now
libpython has the standard install_name relative to rpaths, and
downstream binaries need to add a rpath to the toolchain's lib directory
to load it. This is also now the same behavior as the linux toolchain


jjhelmus:
This change makes sense to me. Thanks @keith

PR #916, opened 2025-12-09
Use latest LLVM for musl builds

### LINKED ISSUE #829: Upgrade musl builds to latest LLVM

jjhelmus:
closes #829

### DIFF
--- a/cpython-unix/build-tcl.sh
+++ b/cpython-unix/build-tcl.sh
@@ -19,12 +19,14 @@ export PKG_CONFIG_PATH=${TOOLS_PATH}/deps/share/pkgconfig:${TOOLS_PATH}/deps/lib
 tar -xf tcl${TCL_VERSION}-src.tar.gz
 pushd tcl${TCL_VERSION}
 
+EXTRA_CONFIGURE=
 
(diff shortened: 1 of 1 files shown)


jjhelmus:
To expand a bit on this change. When compiling for musl with a static option the check for a [proper strtoul implementation](https://github.com/tcltk/tcl/blob/core-8-6-14/unix/configure#L15488) short circuits to `unknown` because of the [cross compiling status](https://github.com/tcltk/tcl/blob/core-8-6-14/unix/configure#L15492-L15493) (building on gnu, target is musl). This results in the substitute [`strtoul` function](https://github.com/tcltk/tcl/blob/core-8-6-14/compat/strtoul.c) being [included in the build](https://github.com/tcltk/tcl/blob/core-8-6-14/unix/configure#L15546). This in turn causes a duplicate definition error because `strtoul` is already provided by libc.

PR #919, opened 2025-12-10
Add support for building with Visual Studio 2026

### LINKED ISSUE #894: Support building on Windows with Visual Studio 2026
Visual Studio 2026 was released in November 2025. The current build setup only supports Visual Studio 2022 and 2019. It would be good to support 2026

jjhelmus:
Add support for building using Visual Studio 2026. This does not change the VS version used to produce the release artifacts only adds support for building with the latest locally. 

closes #894

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -243,6 +243,8 @@ def find_vs_path(path, msvc_version):
         version = "[16,17)"
     elif msvc_version == "2022":
         version = "[17,18)"
+    elif msvc_version == "2026":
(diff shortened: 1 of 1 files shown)


jjhelmus:
Note that even with this change only Python 3.13+ can be built with VS 2026. The property file for 3.12, 3.11 and 3.10 do not have entries for VS version 18 (e.g. 2026). 

This could be added by cherry-picking https://github.com/python/cpython/pull/138980 to the unsupported Python versions.

PR #920, opened 2025-12-11
rename build-uv.py -> build.py

jjhelmus:
Rename the script used to start a build from `build-uv.py` to `build.py`. 
On all platforms builds can be done using:
```
uv run ./build.py ...
```
Additionally on Linux and macOS `./build.py ...` can be used.


jjhelmus:
I will update the docs with these new instructions in another PR once this is merged.

zanieb (approved):

zanieb:
I'd use `ci:skip` on changes like this to avoid massive CI use for a non-build change

jjhelmus:
> I'd use `ci:skip` on changes like this to avoid massive CI use for a non-build change

Since this script is used by CI I wanted to get at least partial coverage to check that I did not miss a rename.

zanieb:
I think this goes way beyond partial coverage though :D

zanieb:
I wouldn't mind if it didn't cause CI runner contention organization-wide

PR #923, opened 2025-12-16
Make the Unix build truly relocatable as CPython patch

mrmathematica:
See #857

### DIFF
--- a/cpython-unix/build.py
+++ b/cpython-unix/build.py
@@ -682,7 +682,8 @@ def python_build_info(
         }
 
         if info.get("build-mode") == "shared":
-            shared_dir = extra_metadata["python_config_vars"]["DESTSHARED"].strip("/")
+            shared_dir = extra_metadata["python_config_vars"]["DESTSHARED"]
+            shared_dir = 'install/lib' + shared_dir.split('install/lib', 1)[1]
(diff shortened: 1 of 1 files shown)


cpython-unix/build-cpython.sh
```diff
@@ -1038,7 +1050,7 @@ metadata = {
     "python_paths_abstract": sysconfig.get_paths(expand=False),
     "python_exe": "install/bin/python%s%s" % (sysconfig.get_python_version(), sys.abiflags),
     "python_major_minor_version": sysconfig.get_python_version(),
-    "python_stdlib_platform_config": sysconfig.get_config_var("LIBPL").lstrip("/"),
+    "python_stdlib_platform_config": sysconfig.get_config_var("LIBPL"),
```
mrmathematica: As as consequence of this change, "python_stdlib_platform_config" will be absolute path instead of relative path. It looks we'd better adopt this change and change doc, instead of trying to force it to remain relative.

If this chane is adopted, update_sysconfig function from https://github.com/astral-sh/uv/blob/13e7ad62cb0d3ca8bf5eeee407d658d6d887d5f9/crates/uv-python/src/sysconfig/mod.rs#L44 can be removed altogether.

PR #926, opened 2025-12-17
Update to CPython 3.15.0a3

edgarrmondragon:


jjhelmus:
Thanks for the update. There are additional changes that are needed for 3.15.0a3, see #927

edgarrmondragon:
> Thanks for the update. There are additional changes that are needed for 3.15.0a3, see #927

Thanks @jjhelmus!

PR #932, opened 2025-12-24
Compile sqlite with SQLITE_ENABLE_UPDATE_DELETE_LIMIT

alindeman:
I have a project that runs sqlite queries in the form `DELETE FROM X ... LIMIT Y`. When I was using a Debian-provided Python, these queries worked properly. But when I used a Python installed by `uv` (via this project), the queries failed with a syntax error.

I tracked it down to a difference in the compile-time options for sqlite. Debian compiles sqlite3 with `-DSQLITE_ENABLE_UPDATE_DELETE_LIMIT=1` but this project dooes not.

Are you all open to adding this compilation flag?

This option enables an optional ORDER BY and LIMIT clause on UPDATE and DELETE statements: https://sqlite.org/compile.html#enable_update_delete_limit

It is enabled by default in the Debian sqlite3 package: https://sources.debian.org/src/sqlite3/3.46.1-8/debian/rules#L51


alindeman:
Unfortunately I realized this won't work as written because we currently build sqlite3 from the [amalgamation](https://sqlite.org/howtocompile.html) and this is one of the few compile-time options that cannot be built using the stock amalgamation.

I will mull over whether it is worth reworking this build process to build the amalgamation from scratch for this option.

jjhelmus:
@alindeman Sorry for the delay getting back, I was traveling last week. 

Are you still interested in adding this option to the sqlite included with pbs? It seems like a reasonable addition.

alindeman:
@jjhelmus Unfortunately it won't work as written because the current build process uses the sqlite amalgamation. Since this change affects the sqlite grammar, the build process would need to be changed to build from individual source files instead.

PR #946, opened 2026-01-08
include tcl9 files in artifacts

jjhelmus:
These files are needed for `_tkinter`. They were mistakenly removed in #928.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -1282,11 +1282,11 @@ fi
 # And prune libraries we never reference.
 rm -f ${ROOT}/out/python/build/lib/{libdb-6.0,libxcb-*,libX11-xcb}.a
 
-if [ -d "${TOOLS_PATH}/deps/lib/tk9.0" ]; then
+if [ -d "${TOOLS_PATH}/deps/lib/tcl9" ]; then
     # Copy tcl/tk resources needed by tkinter.
-    mkdir ${ROOT}/out/python/install/lib/tk9.0
+    mkdir ${ROOT}/out/python/install/lib/tcl
     # Keep this list in sync with tcl_library_paths.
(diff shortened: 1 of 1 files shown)


jjhelmus:
The diff from the commit prior to #928 makes the overall change clear:
``` diff
❯ git diff de54d2de355998491763f49bd9348d66b2ba6fe9 cpython-unix/build-cpython.sh
...
@@ -1268,11 +1282,11 @@ fi
 # And prune libraries we never reference.
 rm -f ${ROOT}/out/python/build/lib/{libdb-6.0,libxcb-*,libX11-xcb}.a

-if [ -d "${TOOLS_PATH}/deps/lib/tcl8" ]; then
+if [ -d "${TOOLS_PATH}/deps/lib/tcl9" ]; then
     # Copy tcl/tk resources needed by tkinter.
     mkdir ${ROOT}/out/python/install/lib/tcl
     # Keep this list in sync with tcl_library_paths.
-    for source in ${TOOLS_PATH}/deps/lib/{itcl4.2.4,tcl8,tcl8.6,thread2.8.9,tk8.6}; do
+    for source in ${TOOLS_PATH}/deps/lib/{itcl4.3.5,tcl9,tcl9.0,thread3.0.4,tk9.0}; do
         cp -av $source ${ROOT}/out/python/install/lib/
     done
 ```

jjhelmus:
> But why are we so inconsistent on quoting?

The quoting is inconsistent in most of the build scripts. I think it evolved that way, likely from various copying from various sources. Running all of scripts through shellcheck, fixing the issues and adding that as a pre-commit/CI check is a good idea.

EliteTK:
> Running all of scripts through shellcheck, fixing the issues and adding that as a pre-commit/CI check is a good idea.

Decrustifying old shell scripts is something i enjoy (believe it or not), might look at it at some point. Thanks for confirming that it would be a welcome change :)

PR #949, opened 2026-01-14
CPython 3.15.0a4 -> 3.15.0a5

jjhelmus:


jjhelmus:
pgo+lto builds on Linux are failing during PGO:
```
cpython-3.15> /tools/host/bin/ld: /tools/deps/lib/libncursesw.a(lib_color.o)
cpython-3.15> :(.bss
cpython-3.15> +0x
cpython-3.15> 4): multiple definition of `
cpython-3.15> COLORS'; Python/optimizer.o (symbol from plugin):(.text+0x0): first defined here
cpython-3.15> /tools/host/bin/ld: warning: size of symbol `COLORS' changed from 4 in Python/optimizer.o (symbol from plugin) to 80 in /tmp/lto-llvm-27918e.o
cpython-3.15> /tools/host/bin/ld: warning: NOTE: size discrepancies can cause real problems.  Investigation is advised.
cpython-3.15> clang: error: linker command failed with exit code 1 (use -v to see invocation)
cpython-3.15> Makefile:1117: recipe for target 'libpython3.15.so' failed
cpython-3.15> make[2]: *** [libpython3.15.so] Error 1
cpython-3.15> make[2]: Leaving directory '/build/Python-3.15.0a5'
cpython-3.15> make[1]: *** [profile-gen-stamp] Error 2
cpython-3.15> Makefile:944: recipe for target 'profile-gen-stamp' failed
cpython-3.15> make[1]: Leaving directory '/build/Python-3.15.0a5'
cpython-3.15> Makefile:955: recipe for target 'profile-run-stamp' failed
cpython-3.15> make: *** [profile-run-stamp] Error 2
```

Perhaps due to CPython shadowing the ncurses `COLORS` symbols?

jjhelmus:
CPython added a `COLORS` symbol in https://github.com/python/cpython/pull/143809 which shadows the symbol of the same name from ncurses.

zanieb:
Do we need to raise that upstream?

jjhelmus:
> Do we need to raise that upstream?

I'll open an issue upstream. 

I think this is only a problem when using LTO and when statically linking the `_curses` extension module. Building a shared `_curses` module is one option but adding `static` to the `COLOR` variable in CPython should also work.

jjhelmus:
The symbol issue has been address upstream in https://github.com/python/cpython/pull/143846, the patch introduced here can be dropped in the next 3.15 release. I'll add a note in the build script shortly.

PR #954, opened 2026-01-18
Remove unused bzip2 symlinks that fail on Python 3.14+

### LINKED ISSUE #953: Builds with Python 3.14+ host fail with `tarfile.AbsoluteLinkError` during dependency extraction
## Summary
Building distributions with a Python 3.14 host interpreter fails during dependency archive extraction due to a change in Python's tarfile defaults.

Python 3.14 included a breaking change to the default `tarfile.extractall()` filter from `fully_trusted` to `data` per [PEP 706](https://peps.python.org/pep-0706/). The `data` filter rejects symbolic links pointing to absolute paths. The dependency archives built by this project (e.g., bzip2) contain absolute symlinks such as `deps/bin/bzcmp`, which triggers `tarfile.AbsoluteLinkError` when extracted.

## Platform
python-build-standalone: `main` @ 06ceef4f1af95a244b22833e59b1bdba0f92caf3

## Steps To Reproduce

```bash
uv run --isolated -p 3.14 ./build.py
```
Alternatively, with a native 3.14 Python install on the host machine:
```bash
python3.14 ./build-<platform>.py
```

## Error
```
  File ".../cpython-unix/build.py", line 1295, in <module>
    sys.exit(main())
  File ".../cpython-unix/build.py", line 1277, in main
    build_cpython(...)
  File ".../cpython-unix/build.py", line 767, in build_cpython
    build_env.install_artifact_archive(BUILD, p, target_triple, build_options)
  File ".../pythonbuild/buildenv.py", line 192, in install_artifact_archive
    extract_tar_to_directory(p, dest_path)
  File ".../pythonbuild/utils.py", line 394, in extract_tar_to_directory
    tf.extractall(dest)
  ...
tarfile.AbsoluteLinkError: 'deps/bin/bzcmp' is a link to an absolute path
```

shaanmajid:
Python 3.14 changed `tarfile.extractall()` default filter from `fully_trusted` to `data`, which rejects absolute symlinks (PEP 706).

bzip2's Makefile creates symlinks (`bzcmp`, `bzless`, `bzegrep`, `bzfgrep`) with absolute paths to the build directory. These are already broken after extraction and aren't needed; only `libbz2.a` and headers are used.

Fixes #953

### DIFF
--- a/cpython-unix/build-bzip2.sh
+++ b/cpython-unix/build-bzip2.sh
@@ -25,3 +25,11 @@ make -j ${NUM_CPUS} install \
     CFLAGS="${EXTRA_TARGET_CFLAGS} -fPIC" \
     LDFLAGS="${EXTRA_TARGET_LDFLAGS}" \
     PREFIX=${ROOT}/out/tools/deps
+
+# bzip2's Makefile creates these symlinks with absolute paths to the build
+# directory, which break after archive extraction. Only libbz2.a and headers
+# are needed for building CPython - remove the shell utility symlinks.
+rm ${ROOT}/out/tools/deps/bin/bzcmp \
+   ${ROOT}/out/tools/deps/bin/bzless \
+   ${ROOT}/out/tools/deps/bin/bzegrep \
+   ${ROOT}/out/tools/deps/bin/bzfgrep


shaanmajid:
Alternative approach could be to edit `pythonbuild/utils.py::extract_tar_to_directory` to use a less restrictive filter.

jjhelmus:
LGMT, thanks for the fix @shaanmajid.

I [ran into this problem](https://github.com/astral-sh/python-build-standalone/pull/892#issuecomment-3582492495) when the uv bootstrap build process was added in #892 but did not address it or add an issue.

It would be nice is there was a target in the Makefile that would only install the library and header but that does not exist.

PR #957, opened 2026-01-21
Publish version metadata to an external repository

zanieb:
Similar to https://github.com/astral-sh/uv/pull/17648

### DIFF
--- a/.github/workflows/publish-versions.yml
+++ b/.github/workflows/publish-versions.yml
@@ -7,6 +7,17 @@ on:
       tag:
         required: true
         type: string
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to publish (e.g. 20260127)"
+        required: true
+        type: string
+      dry-run:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-versions.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: git clone https://${{ secrets.ASTRAL_VERSIONS_PAT }}@github.com/astral-sh/versions.git astral-versions
+
+      - name: "Update versions"
+        run: cat dist/python-build-standalone.json | uv run astral-versions/scripts/publish-version.py --format json --name python-build-standalone --output astral-versions/v1
+
+      - name: "Commit versions"
+        working-directory: astral-versions
+        run: |
+          git config user.name "astral-versions-bot"
+          git config user.email "176161322+astral-versions-bot@users.noreply.github.com"
```
jjhelmus: Does the numerical prefix here has significance? Would a prefix that helped id the commit as coming from the pbs workflow help debug in the future?

PR #960, opened 2026-01-27
Bump OpenSSL 3.5.4 -> 3.5.5

geofft:


jjhelmus:
macOS failing with:
```
libffi> checking for __attribute__((visibility("hidden")))... yes
libffi> clang: error: unsupported option '-print-multi-os-directory'
libffi> clang: error: no input files
```
Odd as nothing in the toolchain or libffi has changed

geofft:
Nah, that particular error message has been around for a while and is harmless. The actual failure is I forgot to update the file size :)

PR #964, opened 2026-01-31
Bump pip to 26.0

mmangkad:


jjhelmus:
Thanks for the update @mmangkad

Issue #259, opened 2024-05-06
Python unable to see CA certificates in Fedora/Rocky/Alma/RHEL systems


nonell-viavi:
Hi,

I _think_ there is an issue when attempting to use this precompiled binaries on a rhel-based system.
I've read the [quirks](https://github.com/indygreg/python-build-standalone/blob/main/docs/quirks.rst) section, and I haven't seen a reference to this problem, that is why I'm raising an issue.

The reason we care about this funcionality is because we're trying to use this python + poetry to get "portable" dev environments accross a bunch of developers (using [mise](https://mise.jdx.dev/lang/python.html#precompiled-python-binaries)).

# Issue description

The following code reproduces the problem on a rhel-based system:

```bash
docker run -it --rm rockylinux:8 bash -c \
  "dnf update -y && dnf install -y wget ca-certificates && \
  wget https://github.com/indygreg/python-build-standalone/releases/download/20240224/cpython-3.10.13+20240224-x86_64-unknown-linux-gnu-install_only.tar.gz && \
  tar -xzf cpython-3.10.13+20240224-x86_64-unknown-linux-gnu-install_only.tar.gz && \
  ./python/bin/python -c \"import urllib.request; f = urllib.request.urlopen('https://www.python.org'); print(f.read(100))\""
```
<details><summary>Logs</summary>

    Connecting to objects.githubusercontent.com (objects.githubusercontent.com)|185.199.108.133|:443... connected.
    HTTP request sent, awaiting response... 200 OK
    Length: 27409290 (26M) [application/octet-stream]
    Saving to: 'cpython-3.10.13+20240224-x86_64-unknown-linux-gnu-install_only.tar.gz'

    cpython-3.10.13+20240224-x86_64-unknown-linux-gnu-in 100%[=====================================================================================================================>]  26.14M  8.59MB/s    in 3.0s

    2024-05-06 08:57:30 (8.59 MB/s) - 'cpython-3.10.13+20240224-x86_64-unknown-linux-gnu-install_only.tar.gz' saved [27409290/27409290]

    Traceback (most recent call last):
      File "/python/lib/python3.10/urllib/request.py", line 1348, in do_open
        h.request(req.get_method(), req.selector, req.data, headers,
      File "/python/lib/python3.10/http/client.py", line 1283, in request
        self._send_request(method, url, body, headers, encode_chunked)
      File "/python/lib/python3.10/http/client.py", line 1329, in _send_request
[...]

PR #976, opened 2026-02-10
Support /etc/pki/tls cert store for older RHEL/Fedora systems

### LINKED ISSUE #858: Look for older RHEL/Fedora /etc/pki/tls cert store location
RHEL 8 (supported until 2029) and below, including Fedora 33 and below, do not ship an `/etc/ssl/cert.pem` or a hashed `/etc/ssl/cert/` directory, so our build of OpenSSL does not pick up platform certs. It would be nice if we patched our Linux build to look at `/etc/pki/tls/cert.pem` instead, if that file exists and `/etc/ssl/cert.pem` does not. Preferably this should be in OpenSSL, but if that's gnarly we can probably get away making this change in the `_ssl` stdlib module.

You can test this with the `redhat/ubi8` or `fedora:33` Docker images, and something like `import urllib.request; urllib.request.urlopen("https://astral.sh")`.

See astral-sh/uv#16703 and https://bugzilla.redhat.com/show_bug.cgi?id=1053882 for more discussion. (I also saw something somewhere about p11-kit having an OpenSSL engine or provider plugin to load the certs from the proper place in the proper way, which we could explore, but seems much more fragile, and would probably also involve figuring out where and how to find the plugin.)

mrmathematica:
Look for /etc/pki/tls/cert.pem when /etc/ssl/cert.pem is not available on RHEL 8 and Fedora 33 and below.

Fixes #858, #259

### DIFF
--- a/cpython-unix/patch-cpython-redhat-cert-file.patch
+++ b/cpython-unix/patch-cpython-redhat-cert-file.patch
@@ -1,16 +1,17 @@
 diff --git a/Lib/ssl.py b/Lib/ssl.py
-index 42ebb8ed384..5cf2575ecd8 100644
+index 42ebb8ed384..c15c0ec940f 100644
 --- a/Lib/ssl.py
 +++ b/Lib/ssl.py
-@@ -423,6 +423,7 @@ class SSLContext(_SSLContext):
+@@ -423,6 +423,8 @@ class SSLContext(_SSLContext):
      """An SSLContext holds various SSL-related configuration options and
      data, such as certificates and possibly a private key."""
      _windows_cert_stores = ("CA", "ROOT")
 +    _FALLBACK_CERT_FILE = "/etc/pki/tls/cert.pem"  # RHEL 8 and below, Fedora 33 and below
++    _FALLBACK_CERT_DIR = "/etc/pki/tls/certs"  # RHEL 8 and below, Fedora 33 and below
  
      sslsocket_class = None  # SSLSocket is assigned later.
      sslobject_class = None  # SSLObject is assigned later.
-@@ -531,6 +532,12 @@ def load_default_certs(self, purpose=Purpose.SERVER_AUTH):
+@@ -531,6 +533,16 @@ def load_default_certs(self, purpose=Purpose.SERVER_AUTH):
          if sys.platform == "win32":
              for storename in self._windows_cert_stores:
                  self._load_windows_store_certs(storename, purpose)
@@ -20,6 +21,10 @@ index 42ebb8ed384..5cf2575ecd8 100644
 +                not os.path.isfile(_def_paths[1]) and
 +                os.path.isfile(self._FALLBACK_CERT_FILE)):
 +                self.load_verify_locations(cafile=self._FALLBACK_CERT_FILE)
++            if (_def_paths[2] not in os.environ and
++                not os.path.isdir(_def_paths[3]) and
++                os.path.isdir(self._FALLBACK_CERT_DIR)):
++                self.load_verify_locations(capath=self._FALLBACK_CERT_DIR)
          self.set_default_verify_paths()
  
      if hasattr(_SSLContext, 'minimum_version'):


jjhelmus:
LGTM. Thanks for putting this together. I want to build this locally and check on an effected system before merging. Should have a result later today.

jjhelmus:
I built Python 3.14 with this patch locally and tested in a Fedora 33 Docker container. 

I am not seeing the cert bundle in /etc/pki/tls/cert.pem being used without setting the `SSL_CERT_FILE` environment variable:

```
./python/install/bin/python -c "import urllib.request; urllib.request.urlopen('https://google.com')"
...
    raise URLError(err)
urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1081)>
```

jjhelmus:
Note that the new location is getting picked up from the patch:

```
./python/install/bin/python -c "import _ssl; print(_ssl.get_default_verify_paths())"
('SSL_CERT_FILE', '/etc/pki/tls/cert.pem', 'SSL_CERT_DIR', '/etc/ssl/certs')
```

mrmathematica:
Thanks @jjhelmus

I see it needs to call `context.load_verify_locations(cafile=SSL_CERT_FILE)` in cpython/Lib/ssl.py to make it work. With this change, now I can verify that `urllib.request` works fine.

jjhelmus:
I think `SSLContext.load_default_certs` may be a better location for a patch. It should be possible to only patch this location and avoid patching `_ssl` entirely.

The patch should check for a Linux platform and then for `/etc/ssl/cert.pem`. If the file does not exist then either the environment variables of the the RHEL default cert file/directory can be loaded.

jjhelmus:
The behavior of the environment variables `SSL_CERT_FILE` and `SSL_CERT_DIR` need to be preserved. When set these variables override the default locations.

mrmathematica:
@jjhelmus  Thanks for the suggestion. Now I see patching `SSLContext.load_default_certs` is a better approach. PR updated.

Environment variables SSL_CERT_FILE and SSL_CERT_DIR was always preserved, previously via `get_default_verify_paths` function.

jjhelmus:
With this change setting `SSL_CERT_FILE` to a path that does not exist will cause the RHEL location to be used. The expected behavior is to use the environment variable location (and likely fail).

mrmathematica:
I see 5 tests failed, all with "Getting latest version from GitHub API...  Error: Github API request failed while getting latest release." Should all be false positives.

mrmathematica:
@jjhelmus   let me know if you have more concerns.

mrmathematica:
> I see 5 tests failed, all with "Getting latest version from GitHub API... Error: Github API request failed while getting latest release." Should all be false positives.

@jjhelmus I see tests got rerun. Now only one test failed, and is the same false positive.

jjhelmus:
This patch should also set the RHEL/Fedora certificate directory (`/etc/pki/tls/certs`) when the `SSL_CERT_DIR` environment variable is not set. 
@mrmathematica If you want to add this it would be great, if not I can do it in a follow on PR.

mrmathematica:
@jjhelmus  I added certificate directory (/etc/pki/tls/certs) as well, thought it is for code completeness only.

nanbiz:
Sorry for being annoying, but do you have an idea when this will be released? I'm thinking of adjusting our CI, but would wait a few days if you said soon™

jjhelmus:
Thanks for putting this together @mrmathematica. Sorry for the delay in merging.

jjhelmus:
> Sorry for being annoying, but do you have an idea when this will be released? I'm thinking of adjusting our CI, but would wait a few days if you said soon™

This change is merged. It will not be available in `uv` until the next release which will be around March 7 (the expected date of the next 3.15 alpha) if not earlier.

PR #979, opened 2026-02-11
update docs with information about building with build.py

jjhelmus:
Document uv based workflow that uses `build.py`.
Remove reference to `build-linux.py`, `build-macos.py` and `build-windows.py` for future removal.

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -4,79 +4,81 @@
 Building
 ========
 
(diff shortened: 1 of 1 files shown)


jjhelmus:
https://github.com/jjhelmus/python-build-standalone/blob/4970684a3c281f96a5f592c2a198fe661a150820/docs/building.rst for a rendered view

PR #981, opened 2026-02-11
CPython 3.15.0a5 -> 3.15.0a6

jjhelmus:
* Update CPython to 3.15.0a6
* Remove unneeded patch added for 3.15.0a5
* Enable tail call for free-threading build on Windows.
* Build `_testinternalcapi` as a shared module.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -325,6 +325,8 @@ if [ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_14}" ]; then
 fi
 
 # Mark _Py_jit_entry as extern in _testiternalcapi/interpreter.c to avoid a duplicate symbols.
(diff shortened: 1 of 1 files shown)


jjhelmus:
macOS error is:
```
cpython-3.15> ld64.lld: error: duplicate symbol: _Py_jit_entry
cpython-3.15> >>> defined in Python/ceval.o
cpython-3.15> >>> defined in Modules/_testinternalcapi/interpreter.o
```

I think this is a result of the change from python/cpython#142911. Splitting out the `_testinternalcapi` module into a extension may address this.

This only happens with optimized builds (pgo, pgo+lto) when `--enable-exterimental-jit` is enabled.

jjhelmus:
`_Py_jit_entry` was added to `Modules/_testinternalcapi/interpreter.c‎` in python/cpython#143944. This should be `extern` when the module is statically linked (and when dynamically linked I think).

PR #984, opened 2026-02-20
[CI] Split build dependencies into own job on Windows

jjhelmus:


jjhelmus:
This would be more useful if a matrix jobs could depend on a specific matrix job rather than the entire jobs to finish. In this way to x86-64+PGO Python builds could start as soon as the x86-64+PGO build-deps job finished rather than waiting for all matrix elements of `build-deps` to finish. This does not seem to be supported in GitHub Action but has been requested, https://github.com/orgs/community/discussions/42335.

zanieb:
> This would be more useful if a matrix jobs could depend on a specific matrix job rather than the entire jobs to finish. In this way to x86-64+PGO Python builds could start as soon as the x86-64+PGO build-deps job finished rather than waiting for all matrix elements of `build-deps` to finish. This does not seem to be supported in GitHub Action but has been requested, https://github.com/orgs/community/discussions/42335.

Hm that's indeed annoying. I guess you could also generate the workflow file.

jjhelmus:
Closing for the time being. May revisit a similar approach in the future.

PR #987, opened 2026-02-27
Fix backticks and unquoted vars in all the bash scripts

EliteTK:
There might be some more candidates but I think the next goal will be to use arrays for all the extra args things to avoid needing to use splitting anywhere, then it should be easy to clean up the rest.

### DIFF
--- a/cpython-unix/build-autoconf.sh
+++ b/cpython-unix/build-autoconf.sh
@@ -5,17 +5,17 @@
 
 set -ex
 
-ROOT=`pwd`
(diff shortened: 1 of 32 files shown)


EliteTK:
I gave it another visual pass and I can't see any case where I only half quote something or where I quoted a glob or something. So I don't think the failures are related, but let me know.

jjhelmus:
The failures look to be unrelated, premature end of the docker images and similar. I re-ran the failed jobs to see if the issue resolves itself.

PR #997, opened 2026-03-04
enable frame pointers for linux x86-64 and aarch64

### LINKED ISSUE #992: Enable frame pointers on Linux for the perf profiler
Python supports tracing with Linux's perf, and ["for best results"](https://docs.python.org/3/howto/perf_profiling.html#how-to-obtain-the-best-results):

> Python should be compiled with `CFLAGS="-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"`

The first of these options was previously discussed in #239 which stalled due to Python deprecating dttace support, but perf support is here to stay.

jjhelmus:
Enable frame pointers on x86-64 and aarch64 Linux platforms.

closes #992

### DIFF
--- a/cpython-unix/targets.yml
+++ b/cpython-unix/targets.yml
@@ -131,6 +131,9 @@ aarch64-unknown-linux-gnu:
   target_cxx: clang++
   target_cflags:
     - '-fvisibility=hidden'
+    # Enable frame pointers
+    - '-fno-omit-frame-pointer'
+    - '-mno-omit-leaf-frame-pointer'
     # Needed to prevent BOLT from crashing.
     - '-fdebug-default-version=4'
   needs:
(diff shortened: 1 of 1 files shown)


jjhelmus:
The native 3.12 build on linux segfaults

> 2026-03-04T18:01:02.9713076Z cpython-3.12> /build/build-cpython.sh: line 889: 11960 Segmentation fault      (core dumped) ${BUILD_PYTHON} "${PIP_WHEEL}/pip" install --prefix="${ROOT}/out/python/install" --no-cache-dir --no-index "${PIP_WHEEL}"

Strange

jjhelmus:
I'm comparing the performance building with frame-pointers vs without to get an idea of how much overhead they introduce. Should have full results later today or tomorrow.

jjhelmus:
Results from running the pyperformance benchmark suite on builds with and without frame-pointer on a Linux x86-64 host:
| Python Version | w-FP geometric mean |
| - | - |
| 3.10 | 1.02x slower |
| 3.11 | 1.02x slower |
| 3.12 | 1.02x slower |
| 3.13 | 1.02x faster |
| 3.14 | 1.03x slower |

Yes, adding frame-pointer improved the performance of 3.13. I don't understand precisely but re-running the benchmarks replicated the results.

Overall a 2-3% performance loss from adding frame-pointers is a reasonable trade-off for the improvement in observability.

PR #1004, opened 2026-03-10
Initialize crypto provider explicitly

zsol:
This was left out of #994 by yours truly


jjhelmus:
The `fetch-release-distributions` runs locally on my mac book with this change.

zsol:
Thanks for validating and sorry for the breakage!

PR #1008, opened 2026-03-13
LLVM toolchain 21.1.4 -> 22.1.1

indygreg:
Let's stay modern.

I performed no verification beyond CI passing.


jjhelmus:
Thanks! I'll run some verifications on the results before merging.

jjhelmus:
I checked interpreters build with LLVM 21.1.1 on Linux and macOS. I did not find anything unusual.

Interestingly this does seem to fix the segfault that occurs when Python 3.12 is built with frame-pointers (#997).

PR #1018, opened 2026-03-20
unix: use extended PGO training set

indygreg:
This will run all tests by default and ensure we have maximal training coverage for PGO.


geofft:
I suspect this PR is equivalent to "get the full test suite passing," which we've talked about wanting to do but hasn't quite been prioritized :)

indygreg:
Yeah. I just wanted to throw up the PR to see what kind of damage we were looking at for enabling the full test harness. It looks substantial :/

PR #1022, opened 2026-03-21
PGO Tweaks

indygreg:
See commit messages and inline comments for more details.

Some of these are good candidates for up-streaming. Although we make assumptions about availability of [modern] LLVM. So upstream patches may require configure work to sniff toolchain capabilities. It was vastly easier to implement the simple patches in PBS first.


geofft (approved):

zanieb:
Can we create some issues for upstreaming per #886

PR #1023, opened 2026-03-21
ci: use larger Depot Linux runners

indygreg:
Default runners have 2 CPUs. We can achieve a nice speed-up (but not quite linear) by leveraging runners with more CPUs.

### DIFF
--- a/ci-runners.yaml
+++ b/ci-runners.yaml
@@ -1,11 +1,11 @@
 # Describes the runners that the CI system can use
 
-depot-ubuntu-22.04:
+depot-ubuntu-22.04-4:
   arch: x86_64
   platform: linux
   free: false
 
-depot-ubuntu-22.04-arm:
+depot-ubuntu-22.04-arm-4:
   arch: aarch64
   platform: linux
   free: false


indygreg:
With the 4 CPU runners (sample size of 1):

* Before: 40m wall; 8971m machine time
* After: 29m wall; 5318m machine time

Perfect scaling of machine time would be 4486m. Since cost scales linearly with CPU count, this change will increase costs by ~18%.

#1022 will help with CPU scaling by making PGO+BOLT run tests in parallel. So probably looking at a 10-15% cost premium after that merges.

IMO this change is worth it since people time costs more than machine time and faster CI enables people to be more productive.

I haven't tested 8 CPU runners thoroughly. But they do yield additional speed-ups.

geofft:
I defer to @zanieb who knows more about our Depot setup and costs, but I would say yes, the turnaround time for getting a Python release out is very important and currently that requires _two_ CI runs (on PR and on HEAD) so paying to get that cut down seems valuable.

(This is probably overengineering but I would likely be in favor of using even larger runners conditionally if the commit bumps a CPython patch version, for the same reason.)

PR #1025, opened 2026-03-22
unix: build test extensions as shared libraries; enable _testcapi

indygreg:
Way back when we only supported statically linked extension modules. And we had to disable `_testcapi` and `_testlimitedcapi` because it wouldn't build statically.

Now that we support building extension modules as shared libraries, we can enable `_testcapi` and `_testlimitedcapi` as shared libraries.

This commit makes that change.

By making this change, we eliminate the largest single source of test failures. But this impact is not captured in code or CI since we currently don't run the stdlib tests in CI.

Enabling the new extensions allows tests - including PGO training tests - to exercise new code paths. This resulted in a new segfault appearing in test_bytes. Why, I'm not sure. (My best guess is it has something to do with calling variadic functions via ctypes.) This particular crash was likely a latent issue. We ignore this crashing test during PGO training as a workaround.

In order to get the extensions enabled on musl shared builds - but not static - we had to introduce some new logic to conditionally enable a shared extension on supported targets while automatically disabling on static targets. There may be a cleaner way to express this semantic intent. But the implemented solution does work.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -616,6 +616,11 @@ if [[ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_14}" ]]; then
     PROFILE_TASK="${PROFILE_TASK} --ignore test_json"
 fi
 
+# PGO optimized / BOLT instrumented binaries segfault in a test_bytes test. Skip it.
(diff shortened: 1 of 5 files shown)


indygreg:
I'm going to continue iterating on this to get musl shared builds to behave better.

indygreg:
> Seems good. The only thing I'm wondering is whether we want to ship these test extensions in the install-only-stripped build or drop them—most users will not need them.

These extensions are only used by the stdlib tests AFAICT. So it makes sense to strip them from the install-only artifacts.

PR #1026, opened 2026-03-23
ci: elide apt package install if packages present

indygreg:
At least on the Depot runners these packages are already present. CI is spending ~30s upgrading the packages to no benefit. This is contributing to the long pole of CI turnaround times and slowing down feedback on CI.

Let's elide this overhead if possible.


indygreg:
Bleh. pkg-config isn't installed on the x86-64 runners.

PR #998, opened 2026-03-04
add deprecation warnings to build-{linux,macos}.py

jjhelmus:
Add a deprecation warning to the build-linux.py and build.macos.py scripts informing users of the their future removal and directing users to ./build.py.

PR #1029, opened 2026-03-23
remove deprecated build-{linux,macos,windows}.py scripts

jjhelmus:
Remove the deprecated build-linux.py, build-macos.py and build-windows.py scripts and associated files.

Build should be started using `uv run build.py` or `./build.py`.

The build-linux.py and build-windows.py have shown a warning about future removal since #998. 
The removal of these scripts was mentioned in the release notes of 20260310 and 20260320


jjhelmus:
I have no concerns about delaying these removal of these scripts, especially `build-windows.py` which is lacking a warning message and notice about removal.

Did want to see if these is anything that still uses this, like the check action in CI.

jjhelmus:
#1040 migrates check.py to uv from the pip created venv. 

This PR requires that to be merged first.

Issue #721, opened 2025-08-03
Do not statically link _ctypes_test module


geofft:
It (intentionally) exports a bunch of C symbols with unprefixed names like `left`, `right`, `top`, `bottom`, `myprintf`, `an_integer`, `getSPAMANDEGGS`, etc. We shouldn't be exporting these symbols in `bin/python` or `libpython`.

We probably want to keep building it for the sake of running the tests that use it, we just want to build it dynamically.

It might be cool to add a test at the same time that all our exported dynamic symbols start with `Py` or `_Py`. There is some stuff in src/validation.rs to check what we're doing with symbols but I guess it doesn't check this.

PR #1032, opened 2026-03-24
release: remove test extensions from install-only archives

indygreg:
These extensions are only needed to support the stdlib tests, which we strip from these archives. These extensions offer no utility. So don't ship them.

This is a follow-up to #1025. Implementing as its own PR since we remove previously existing extension modules. This is backwards incompatible and deserves to be discussed on its own merits.

### DIFF
--- a/src/macho.rs
+++ b/src/macho.rs
@@ -4,7 +4,7 @@
 
 use {
     crate::validation::ValidationContext,
-    anyhow::{Context, Result, anyhow},
+    anyhow::{anyhow, Context, Result},
     apple_sdk::{AppleSdk, SdkSearch, SdkSearchLocation, SdkSorting, SdkVersion, SimpleSdk},
     semver::Version,
     std::{
--- a/src/release.rs
+++ b/src/release.rs
@@ -448,6 +448,19 @@ pub fn build_wanted_filenames(
     Ok(wanted_filenames)
 }
 
+/// Extension modules that should not be included in "install only" archives.
+const INSTALL_ONLY_DROP_EXTENSIONS: &[&str] = &[
+    "_ctypes_test",
+    "_testbuffer",
+    "_testcapi",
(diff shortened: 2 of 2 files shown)


geofft:
The breaking change is the unshipping of `_ctypes_test`, right? That's a very goofy extension that I think has only ever caused our users problems.

indygreg:
> The breaking change is the unshipping of `_ctypes_test`, right? That's a very goofy extension that I think has only ever caused our users problems.

And unshipping other `_test*` extensions (like `_testbuffer`). #1025 is only enabling `_testcapi` and `_testlimitedcapi`.

I'm curious how `_ctypes_test` is causing problems for your users. It was made to build as a shared module in #747. Presumably it is no longer getting imported by anything and not causing the symbol conflicts detailed by #721.

geofft:
#721 is what I meant - I cannot remember `_ctypes_test` ever doing anything good, only doing bad things. :) But yeah, I see, the other `_test*` modules were statically linked.

There's a sort of related question here about whether to ship the Python parts of the test suite (the `test` module), and I kind of think the answer is we shouldn't in install-only builds. That would also be a breaking change, so maybe there's an argument to do them both in the same release.

FWIW, I'm okay with this personally. On the python-build-standalone side, I don't think there's anything else pending we want to get out the door in the next few days, so we can cut a release with this change alone (maybe plus any PGO stuff you feel is ready). On the uv side, we did _just_ get a breaking uv version bump (0.11.0) out the door today so we're unlikely to do another one any time soon, but we usually don't treat potentially-breaking p-b-s changes as breaking changes in uv, and again this one seems so unlikely to impact real users, so I don't think that should influence anything. But I'll defer to @zanieb and @jjhelmus if they have thoughts.

zsol:
Out of curiosity, how much do these contribute to the artifact size?

PR #1037, opened 2026-03-24
ci: optimize Docker image loading

indygreg:
Loading images step currently takes ~30s +- a ton of variance.

Linux CI jobs are currently taking 8-20m typically. So image builds can be up to ~6% of wall time.

I've been playing around with further increasing Depot runner sizes. That delivers a considerable wall time reduction (<5m builds possible). This increases the percentage overhead of any "CI taxes," like restoring caches and pulling/loading container images.

This commit implements a pair of optimizations to enable the image loading step to take as little as 5s:

* We parallelize the image decompression + loading using simple `&` + `wait` shell features.
* We filter out images that aren't needed for the current job.

The filtering isn't fully lean - we don't currently filter out images related to cross-compiling.

And we still incur overhead to download images we don't load.

Those optimizations will wait for another day.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -2,7 +2,7 @@ name: linux
 
 on:
   push:
-    branches: [main]
+    branches: [ main ]
   pull_request:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


jjhelmus:
If you are looking at speeding up CI, splitting building dependencies from cpython and then reusing the built dependencies has the potential for a good improvement. #984 has a start for Windows.

PR #1052, opened 2026-03-25
CI: use uv instead of pip

zsol:
This introduced a bit of a circular dependency, but as long as we use a pinned uv version it should be fine.


jjhelmus:
This is similar to #1040. 
The main difference is the dependency group. 1040 adds the requirements in a `check` group to avoid introducing those requirements into the environment used for the build.

zsol:
Oh nice! I'll rebase the other pr on top of that then.

PR #1055, opened 2026-03-25
Make dry runs exercise the mirror publishing code

zsol:
To achieve this, the upload command now writes out the `SHA256SUMS` file _before_ uploading them to github. During dry-run mode, the upload command exits immediately after writing the checksums.

This is because the dry-run of the mirror upload step needs an existing `SHA256SUMS` file.

## Tests

A new test was added to prove that dry-run mode still doesn't talk to GitHub (but does write the checksum file).

Example dry run: https://github.com/astral-sh/python-build-standalone/actions/runs/23662482434

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,7 +101,6 @@ jobs:
             dist/*.tar.zst
 
       - name: Publish to Astral mirror
-        if: ${{ github.event.inputs.dry-run == 'false' }}
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.MIRROR_R2_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.MIRROR_R2_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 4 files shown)


jjhelmus:
Having build.py use `--no-dev` gives me pause is it is the primary entry point of the `python-build-standalone` package. On the other hand it is natural to add boto3 as a requirement and moto as a dev requirement when looking at `pythonbuild` as a library.

I don't have strong feeling one way or the other. Keep the setup here or move moto to a different dependency group and us the pattern in check.py.

zanieb:
Using `--no-dev` everywhere seems pretty meh. I'd add a different dependency group.

indygreg:
pythonbuild is a local only package to support PBS development. Everything is an implicit "dev" dependency. Unless I'm missing something, dependency groups feels like avoidable complexity.

jjhelmus:
Dependency groups allow for a different and typically smaller environments to be used for the different scripts. For example `build.py` does not need ruff or mypy which are needed to run `check.py`. But this also increase complexity, requiring coordinating which scripts use which environments. 

Given that there has not been any issues around the python environment size or time to create it I'd be in favor of just dumping everything into requirements and revisit is issues come up.

zsol:
Apologies, y'all were commenting on https://github.com/astral-sh/python-build-standalone/pull/1051, which this PR was stacked on top of, but when the former was merged, I forgot to rebase this.

PR #1056, opened 2026-03-26
unix: support building Linux x86-64 from macOS aarch64

indygreg:
Docker on aarch64 macOS will automagically virtualize x86-64 containers if containers are spawned with `platform=linux/amd64`. Performance of spawned containers is a bit slower than native, but not horrible. This functionality means it is viable to develop Linux x86-64 from modern Apple hardware.

This commit teaches the build system to support cross-compiling Linux x86-64 from macOS aarch64.

Implementing this wasn't too difficult: we need to pass `platform` into Docker's APIs for building and creating containers. We need to teach code to resolve the effective host platform when this scenario is detected. And we need to advertise support for cross-compiling in the `targets.yml` file.

In case you are wondering, yes, a similar solution could be employed for Linux too by using emulation. But this requires Docker be configured to support emulation, which isn't as common. Rosetta on macOS "just works" and is therefore the lowest hanging fruit to implement.

### DIFF
--- a/cpython-unix/build-main.py
+++ b/cpython-unix/build-main.py
@@ -142,7 +142,14 @@ def main():
     effective_host_platform = host_platform
     if building_linux_from_macos:
         if host_platform == "macos_arm64":
-            effective_host_platform = "linux_aarch64"
+            if target_triple.startswith("aarch64"):
+                effective_host_platform = "linux_aarch64"
+            elif target_triple.startswith("x86_64"):
+                effective_host_platform = "linux_x86_64"
+            else:
+                raise Exception(
+                    f"unsupported macOS cross-compile: {host_platform} -> {target_triple}"
+                )
         else:
             raise Exception(f"Unhandled macOS platform: {host_platform}")
         print(
--- a/cpython-unix/build.py
+++ b/cpython-unix/build.py
@@ -25,7 +25,12 @@ from pythonbuild.cpython import (
     meets_python_minimum_version,
     parse_setup_line,
 )
-from pythonbuild.docker import build_docker_image, get_image, write_dockerfiles
+from pythonbuild.docker import (
+    build_docker_image,
+    docker_platform_from_host_platform,
+    get_image,
+    write_dockerfiles,
+)
 from pythonbuild.downloads import DOWNLOADS
 from pythonbuild.logging import log, set_logger
 from pythonbuild.utils import (
@@ -33,6 +38,7 @@ from pythonbuild.utils import (
     add_licenses_to_extension_entry,
     clang_toolchain,
     create_tar_from_directory,
+    current_host_platform,
     download_entry,
     get_target_settings,
     get_targets,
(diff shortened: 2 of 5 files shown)


jjhelmus:
I tried this out locally and after building for `aarch64-unknown-linux-gnu` and removing the build directory a follow up build for `x86_64-unknown-linux-gnu` fails when building the image-gcc target. 

For some reason armhf deb packages are being installed in the container and `uname -a` reports `Linux 9ceedc289281 6.10.14-linuxkit #1 SMP Tue Oct 14 07:32:13 UTC 2025 armv7l GNU/Linux`.

I think this is caused by docker-py not fully respecting the platform argument, see docker/docker-py#2990. There may be a workaround that involves removing cached images.

indygreg:
That's a fun quirk. I'll take a look and try to concoct a workaround. I know a bit about Docker/Buildkit internals and may be able to devise something non-obvious.

indygreg:
Yeah, I bet the code in `docker buildx build` handles this platform issue. Whereas the raw Docker image build API (which abstracts away the BuildKit bits) doesn't. We may want to switch to the CLI command for image building. I may do that as its own PR.

indygreg:
Oh, I think it is sillier than that. We're adding `platform` when cross compiling on Linux. That forces Docker to run an image from a non-native arch, invalidating assumptions that Linux images are always for the native arch.

jjhelmus:
I'm still getting armhf Debian packages installed into the `image-build` and `image-gcc` containers when building with `--target-triple x86_64-unknown-linux-gnu` on a macOS host. I may have some armhf debian containers in my cache from debugging issues with that platform but the cache should not be effecting the platforms selection.

indygreg:
> I'm still getting armhf Debian packages installed into the `image-build` and `image-gcc` containers when building with `--target-triple x86_64-unknown-linux-gnu` on a macOS host. I may have some armhf debian containers in my cache from debugging issues with that platform but the cache should not be effecting the platforms selection.

Right - our build system is supposed to load the local image tags into Docker and then execute those images directly by specifying their SHA-256 digest. The only time we would be loading incorrect images is if the `docker build` equivalent that we're running is producing an _invalid_ image and we're loading that.

Try removing `build/image-*.tar` files and see if you can reproduce?

jjhelmus:
I tracked down why the wrong platform was being used on my Mac. client.api.build() uses Docker’s legacy builder rather than BuildKit. As described in docker/docker-py#2990, the legacy builder can incorrectly reuse a cached base-image variant despite an explicit platform argument. This can cause an image for the wrong architecture to be produced.

A small reproducer is:
``` Python
#!/usr/bin/env -S uv run --no-dev
import io
import docker

IMAGE = "debian@sha256:32ad5050caffb2c7e969dac873bce2c370015c2256ff984b70c1c08b3a2816a0"
client = docker.from_env(timeout=600)

print("Seeding the local cache with linux/arm/v7...")
for _ in client.api.pull(IMAGE, platform="linux/arm/v7", stream=True, decode=True): pass

print("Requesting linux/amd64 through docker-py...")
fh = io.BytesIO(f"FROM {IMAGE}\nRUN uname -m && dpkg --print-architecture\n".encode())
for event in client.api.build(fileobj=fh, platform="linux/amd64", decode=True):
    if "stream" in event:
        print(event["stream"], end="")
 ```
 On my machine, this prints armv7l and armhf despite requesting linux/amd64.
 
A workaround is to invoke docker buildx build in a subprocess so the image is built with BuildKit, which correctly resolves the requested platform. I’ve applied that workaround on top of this branch and submitted it as #1179.

PR #1058, opened 2026-03-26
Port `verify_distribution.py` into `pythonbuild`

indygreg:
We move the custom `unittest` based tests from this file into a new `pythonbuild/disttests` package, deleting `src/verify_distribution.py` in the process.

We teach the new `pythonbuild.testdist` code to run our custom unittests given an extracted Python distribution. They now run by default.

Rust code for executing the Python in the distribution has been removed.

CI tests have been updated to invoke `test-distribution.py` when we support running the interpreter.

Behavior of `test-distribution.py` / `pythonbuild.disttests` has been changed so stdlib tests are no longer run by default. This is necessary to preserve backwards compatibility to keep CI passing.

Functionality is still a bit klunky IMO. I'll clean up the CLI UX a bit in future commits.


jjhelmus:
The merge conflicts resulting from merging #1057 and #1051 will need to be resolved but then this looks good.

PR #1062, opened 2026-03-27
Revert "Port `verify_distribution.py` into `pythonbuild`"

indygreg:
Reverts astral-sh/python-build-standalone#1058.

This caused Windows ARM bustage due to `cryptography` OpenSSL dependency wonkiness.


indygreg:
#1063 is a better solution.

Issue #956, opened 2026-01-20
Thread creation fails under systemd's `MemoryDenyWriteExecute=true`


bluetech:
systemd has a hardening flag for services `MemoryDenyWriteExecute`, described in the manpage as folllows:

>  Takes a boolean argument. If set, attempts to create memory mappings that are writable and executable at the same time, or to change existing memory mappings to become executable, or mapping shared memory segments as executable, are prohibited. Specifically, a system call filter is added (or preferably, an equivalent kernel check is enabled with prctl(2)) that rejects mmap(2) system calls with both PROT_EXEC and PROT_WRITE set, mprotect(2) or pkey_mprotect(2) system calls with PROT_EXEC set and shmat(2) system calls with SHM_EXEC set. Note that this option is incompatible with programs and libraries that generate program code dynamically at runtime, including JIT execution engines, executable stacks, and code "trampoline" feature of various C compilers. This option improves service security, as it makes harder for software exploits to change running code dynamically. [...]

When using this option with uv's python (3.14, Linux, x86-64, gnu), thread creation from Python fails:

```
$ systemd-run -p MemoryDenyWriteExecute=true -t ~/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/bin/python -c 'import threading; threading.Thread().start()'
Running as unit: run-p344916-i344917.service
Press ^] three times within 1s to disconnect TTY.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import threading; threading.Thread().start()
                      ~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/ran/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/threading.py", line 1004, in start
    _start_joinable_thread(self._bootstrap, handle=self._os_thread_handle,
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                           daemon=self.daemon)
                           ^^^^^^^^^^^^^^^^^^^
RuntimeError: can't start new thread
```

This doesn't happen with system Python (tested on Arch Linux with Python 3.14 and Debian trixie with Python 3.13):

```
$ systemd-run -p MemoryDenyWriteExecute=true -t python -c 'import threading; threading.Thread().start()'
Running as unit: run-p345006-i345007.service
Press ^] three times within 1s to disconnect TTY.
```

[...]

PR #1064, opened 2026-03-27
include noexecstack hardening flag on Linux

### LINKED ISSUE #1061: 20260320 builds seem to have executable-stack set on python shared library
I must admit I don't exactly know what the above means, but I encountered `pyinstaller` errors (copied below) which were traced back to apparent changes in these Python builds:

> My first hypothesis is thus https://github.com/astral-sh/python-build-standalone/releases/tag/20260320. 
> 
> This indeed seems to be the case. 
> 
> The builds in this release seem to have executable-stack set on python shared library, which can refuse to be loaded on sufficiently hardened systems (e.g., by SELinux).
> 
> ```
> $ wget https://github.com/astral-sh/python-build-standalone/releases/download/20260320/cpython-3.14.3+20260320-x86_64-unknown-linux-gnu-install_only.tar.gz
> $ tar xvf cpython-3.14.3+20260320-x86_64-unknown-linux-gnu-install_only.tar.gz
> $ execstack -q python/lib/libpython3.14.so
> X python/lib/libpython3.14.so
> ``` 
> 
> vs. previous release:
> 
> ```
> $ wget https://github.com/astral-sh/python-build-standalone/releases/download/20260310/cpython-3.14.3+20260310-x86_64-unknown-linux-gnu-install_only.tar.gz
> $ tar xvf cpython-3.14.3+20260310-x86_64-unknown-linux-gnu-install_only.tar.gz
> $ execstack -q python/lib/libpython3.14.so
> - python/lib/libpython3.14.so
> ```
> 
> As their `python` interpreter is not linked against the shared library (it's one of those builds where executable is statically linked against python shared library, and the latter is provided only for embedders, such as PyInstaller), it is not directly affected by this issue; and so I imagine it is easy for a problem like this to go unnoticed on their side.
> 
>  

 _Originally posted by @rokm in [#9413](https://github.com/pyinstaller/pyinstaller/issues/9413#issuecomment-4118493151)_

jjhelmus:
Mark the stack memory as non-executable using the '-Wl,-z,noexecstack' flag on aarch64 and x86_64 linux platforms.

Other linux targets are cross-compiled, this flag is left off for the time being.

closes #1061

### DIFF
--- a/cpython-unix/targets.yml
+++ b/cpython-unix/targets.yml
@@ -136,6 +136,9 @@ aarch64-unknown-linux-gnu:
     - '-mno-omit-leaf-frame-pointer'
     # Needed to prevent BOLT from crashing.
     - '-fdebug-default-version=4'
+  target_ldflags:
+    # Hardening
+    - '-Wl,-z,noexecstack'
   needs:
     - autoconf
     - bdb
@@ -575,6 +578,9 @@ x86_64-unknown-linux-gnu:
     - '-mno-omit-leaf-frame-pointer'
     # Needed to prevent BOLT from crashing.
     - '-fdebug-default-version=4'
+  target_ldflags:
+    # Hardening
+    - '-Wl,-z,noexecstack'
   needs:
     - autoconf
     - bdb
@@ -625,6 +631,9 @@ x86_64_v2-unknown-linux-gnu:
     - '-mno-omit-leaf-frame-pointer'
     # Needed to prevent BOLT from crashing.
     - '-fdebug-default-version=4'
+  target_ldflags:
+    # Hardening
+    - '-Wl,-z,noexecstack'
   needs:
     - autoconf
     - bdb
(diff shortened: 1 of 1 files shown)


jjhelmus:
Testing this locally confirms this addresses #1061

```
❯ execstack -q python/install/lib/libpython3.14.so
- python/install/lib/libpython3.14.so
```

This change does not address #956

geofft:
This is fine, but I'd also like to track down _why_ this is happening, if possible. Is it some assembly file somewhere?

jjhelmus:
> This is fine, but I'd also like to track down _why_ this is happening, if possible. Is it some assembly file somewhere?

I should have included a mention of https://github.com/llvm/llvm-project/issues/186004 in this issue and the commit message. It is discussed in #1061 

LLVM 22 does not emit a `.note.GNU-stack` section when `-Wa,--noexecstack` is passed to the assembler. This has been fixed in the dev branch of llvm and I think has been or will be backported to LLVM 22.

geofft:
Wait, why is this not relevant for cross-compiled targets?

jjhelmus:
> Wait, why is this not relevant for cross-compiled targets?

A non-executable stack is relevant for all targets, both native and cross but the cross-compiled targets use gcc for compilation so should not be effected by the llvm bug.

That said, adding `-Wl,-z,noexecstack` to LDFLAGS on all Linux targets is a good idea.

PR #1070, opened 2026-03-29
Validate that ELF objects request a non-executable stack

### LINKED ISSUE #1069: Add validation for none-executable stack
With the changes in #1064 and #1065 both `python` and `libpython` are marked as having a non-executable stack. This characteristic should be validated after the build.

geofft:
Closes #1069.

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -1161,23 +1161,37 @@ fn validate_elf<Elf: FileHeader<Endian = Endianness>>(
         }
     }
 
-    let mut found_pt_gnu_stack = false;
-    for phdr in elf.program_headers(endian, data)? {
-        if phdr.p_type(endian) != PT_GNU_STACK {
-            continue;
+    // Verify that objects are not requesting an executable stack. For backwards compatibility,
+    // Linux (the kernel when loading an executable, and glibc when loading a shared library)
+    // assumes you need an executable stack unless you request otherwise. In linked outputs
+    // (executables and shared libraries) this is in the program header: the flags of a
+    // PT_GNU_STACK entry specify stack permissions, and the default if unspecified is RWX. In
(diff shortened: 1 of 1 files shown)


jjhelmus:
https://github.com/llvm/llvm-project/issues/186004 will cause object files assembled with LLVM 22 to fail this check. 
These could be post-processed with objcopy to add the section. Another option is to limit this check to only the ELF objects in the install directory or potentially just python and libpython.

geofft:
Yeah, I also missed the distinction between executables/libraries which have the PT_GNU_STACK program headers, and intermediate objects that have the section but no program headers (at all). I'll make this check conditional on final objects (ET_EXEC or ET_DYN, I think), and we can defer checking .o files to later.

jjhelmus (approved):

jjhelmus:
LGTM. Should we validate this on other architectures before meting?

geofft:
I was gonna pay attention to CI post-merge and then open a new PR to do `-Wl,-z,noexecstack` if there are any failures. Lemme know if that seems too risky :)

jjhelmus:
Watching the post merge for failures is reasonable. 
Merging. I will keep an eye out for failures. Thanks.

jjhelmus:
riscv64 builds are failing verification.
```
python/install/lib/thread3.0.4/libtcl9thread3.0.4.so missing PT_GNU_STACK header (defaults to executable stack)
```
https://github.com/astral-sh/python-build-standalone/actions/runs/24349377629/job/71099408609

I'm taking a look

jjhelmus:
Confirmed the missing GNU_STACK header in the riscv64 builds, mostly in Tcl/Tk related libraries, locally. This is not new, the header is missing in an older (20251010) release as well.

Adding in an explicit flag to avoid this should be done, if too much work this test can be skipped on the architecture. I'll prepare both after lunch.

PR #1071, opened 2026-03-30
Add regression tests for MemoryDenyWriteExecute=true

geofft:
As reported in #956, executable stacks cause thread creation to fail under MemoryDenyWriteExecute=true. In kernel 6.3+ (which the GitHub Actions runners should have) this is implemented using prctl(PR_SET_MDWE). In older versions systemd uses a more complicated seccomp filter.

### DIFF
--- a/pythonbuild/disttests/__init__.py
+++ b/pythonbuild/disttests/__init__.py
@@ -320,6 +320,42 @@ class TestPythonInterpreter(unittest.TestCase):
         with self.subTest(msg="weird argv[0]"):
             assertPythonWorks(sys.executable, argv0="/dev/null")
 
+    @unittest.skipUnless(sys.platform == "linux", "Linux-specific prctl")
+    @unittest.skipIf(
+        "static" in os.environ["BUILD_OPTIONS"],
+        "cannot import libc on static builds",
+    )
+    def test_nx_thread_creation(self):
+        "Test that thread creation works under e.g. systemd's MemoryDenyWriteExecute."
+        # Note that NX cannot be unset so this pollutes the current process,
+        # but if something else breaks under NX we probably want to know!
+        import ctypes
+        import threading
+
+        libc = ctypes.CDLL(None, use_errno=True)
+        # <linux/prctl.h>
+        PR_SET_MDWE = 65
+        PR_GET_MDWE = 66
+        PR_MDWE_REFUSE_EXEC_GAIN = 1 << 0
+        PR_MDWE_NO_INHERIT = 1 << 1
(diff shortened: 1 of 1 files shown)


geofft:
Fails locally on an existing release, although unfortunately Python's error message is not great:

```
$ BUILD_OPTIONS= TARGET_TRIPLE=x86_64-linux-gnu uvx python3 -m unittest pythonbuild.disttests 
......E..Es.
======================================================================
ERROR: test_nx_thread_creation (pythonbuild.disttests.TestPythonInterpreter.test_nx_thread_creation)
Test that thread creation works under e.g. systemd's MemoryDenyWriteExecute.
----------------------------------------------------------------------
  File "/home/ubuntu/python-build-standalone/pythonbuild/disttests/__init__.py", line 350, in test_nx_thread_creation
    t.start()
    ~~~~~~~^^
  File "/home/ubuntu/.local/share/uv/python/cpython-3.14.3-linux-x86_64-gnu/lib/python3.14/threading.py", line 1005, in start
    _start_joinable_thread(self._bootstrap, handle=self._os_thread_handle,
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                           daemon=self.daemon)
                           ^^^^^^^^^^^^^^^^^^^
RuntimeError: can't start new thread
```

jjhelmus (approved):

pythonbuild/disttests/__init__.py
```diff
@@ -320,6 +320,42 @@ def assertPythonWorks(path: Path, argv0: Optional[str] = None):
...
 
+    @unittest.skipUnless(sys.platform == "linux", "Linux-specific prctl")
+    @unittest.skipIf(
+        "static" in os.environ["BUILD_OPTIONS"],
+        "cannot import libc on static builds",
+    )
+    def test_nx_thread_creation(self):
+        "Test that thread creation works under e.g. systemd's MemoryDenyWriteExecute."
+        # Note that NX cannot be unset so this pollutes the current process,
+        # but if something else breaks under NX we probably want to know!
```
jjhelmus, line 331: Is it worth isolating this from other tests via subprocess? Not a blocker to merging.

jjhelmus (approved):

pythonbuild/disttests/__init__.py
geofft: Is that easy to do? My gut feeling was the implementation complexity of doing that outweighed the potential benefits (especially since other failures under W^X are probably legitimate bugs) but if there's an easy way to do it I'm fine with it.

PR #1075, opened 2026-04-01
Run a smaller subset of targets on pull requests by default

zanieb:
Currently, pull requests run all targets by default and labels can be used to select a subset of the matrix. However, CI is quite expensive and it's very rare to need to test the whole matrix, so the default is a bit backwards.

In this change, we update pull requests to run on a subset of targets by default: Python 3.14 on macOS, Linux, and Windows with the most popular architecture. We include both glibc and musl variants of Linux. We include armv7 as an arbitrary cross-compile case — I'd be happy to take suggestions on an alternative there.

This breaks our labeling concept a bit, as our labels currently do pure subsetting of the matrix. The labels will continue to subset, but with some nuances:

- Applying a label will generally subset the default selection, e.g., `platform:linux` will only run the Linux subset of the defaults described above.
- Labels that would null the default subset, e.g., `python:3.12` or `build:debug`, will instead change the default set to target that variant.
-  There are new `platform:all`, `python:all`, `arch:all`, `libc:all`, and `build:all` labels that can be used to expand the targets.


jjhelmus (approved):

ci-matrix.py
```diff
@@ -75,46 +76,201 @@ def parse_labels(labels: str | None) -> dict[str, set[str]]:
...
+    pattern: dict[str, str | None],
+    expand_platform: bool,
+    expand_arch: bool,
+    expand_libc: bool,
+) -> bool:
+    if not expand_platform and target_platform != pattern["platform"]:
         return False
 
-    if filters.get("build"):
-        build_options = set(entry.get("build_options", "").split("+"))
```
geofft: This logic is lost in the rewrite, i.e., `build:pgo` or `build:static` or `build:freethreaded` no longer works I think.

geofft (commented):
This seems pretty complicated and I'm not sure that I understand exactly what the logic is and a mental model of which builds will run in response to which targets. To be fair I don't feel like I have a 100% grasp of the current logic, and we could decide that we have few enough existing users that backwards compatibility is not a goal. But it feels like it should be possible to do this task with fewer lines of code and more readably.

I'm okay with deciding that we don't want to put time into that, and we can always have an LLM make further big changes if we don't like the behavior.

ci-matrix.py
```diff
@@ -75,46 +76,163 @@ def parse_labels(labels: str | None) -> dict[str, set[str]]:
...
+    allowed_triples: set[str] | None = None,
+) -> bool:
+    """Check if a matrix entry should be included.
+
+    For pull requests, entries are restricted to the allowed target set
+    (computed by expand_default_triples), the default python version, and
+    the curated build options — unless overridden by labels. For pushes
+    (pull_request_defaults is None), only label filters apply.
+    """
+    # PR default restrictions
```
zanieb:
```suggestion
```

ci-matrix.py
```diff
@@ -380,15 +538,62 @@ def main() -> None:
...
+        with open(CI_DEFAULTS_YAML) as f:
+            ci_defaults = yaml.safe_load(f)
+
+        pull_request_defaults = ci_defaults.get("pull_request")
+        if pull_request_defaults is None:
+            print(
+                f"error: {CI_DEFAULTS_YAML} is missing a pull_request section",
+                file=sys.stderr,
+            )
+            sys.exit(1)
```
zanieb: It seems like this should just be unconditionally loaded and we can do `ci_defaults.get(args.event)`?

zanieb: (and not error if there's not a section for the event type, since that just means... use everything)

PR #1096, opened 2026-04-15
Fix 2-factor release process when target commit does not match workflow commit

zanieb:


.github/workflows/release.yml
```diff
@@ -22,19 +22,65 @@ permissions: {}
...
+      - name: Resolve target SHA
+        id: vars
+        env:
+          INPUT_SHA: ${{ github.event.inputs.sha }}
+        run: |
+          target_sha="$INPUT_SHA"
+          if [ -z "$target_sha" ]; then
+            target_sha="$GITHUB_SHA"
+          fi
+          echo "target_sha=$target_sha" >> "$GITHUB_OUTPUT"
```
zanieb, line 47: No clue why Codex wrote this convoluted step

PR #1099, opened 2026-04-17
Run a small subset of targets on pull requests w/o labels

jjhelmus:
Alternative to #1075.

Run a small subset of targets on pull requests when no labels are specified.

If labels are specified they filter the full set of targets.

On a push all targets are run.

Running all targets on a pull request requires using multiple labels, for example `platform:linux`, `platform:darwin`, `platform:windows`.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -178,6 +178,7 @@ jobs:
             --platform linux \
             --labels "${STEPS_GET_LABELS_OUTPUTS_LABELS}" \
             --max-shards 2 \
+            --event ${{ github.event_name }} \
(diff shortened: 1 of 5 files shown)


zanieb:
> If labels are specified they filter the full set of targets.

I don't think this is good enough (which is why I avoided it), e.g., it should be easy to switch Python versions with reasonable defaults for the rest of the target components.

jjhelmus:
> it should be easy to switch Python versions with reasonable defaults for the rest of the target components.

Agreed, being able to change the default set to a specific target (python version, build option) is something I would use often. It is a feature of #1075 that I'm appreciating more as I think about it.

zanieb:
If you're comfortable with it, we can just ship that and iterate on any issues and confusion?

jjhelmus:
closing in favor of #1075

PR #1101, opened 2026-04-21
Cross compile with LLVM and a sysroot for riscv64

jjhelmus:
Use LLVM 22 and a Debian buster sysroot to cross-compile for riscv64.

### DIFF
--- a/cpython-unix/build.cross-riscv64.Dockerfile
+++ b/cpython-unix/build.cross-riscv64.Dockerfile
@@ -1,6 +1,6 @@
-# Debian Buster.
-FROM debian@sha256:2a0c1b9175adf759420fe0fbd7f5b449038319171eb76554bb76cbe172b62b42
-LABEL org.opencontainers.image.authors="Gregory Szorc <gregory.szorc@gmail.com>"
+# Debian Trixie
+FROM debian@sha256:3352c2e13876c8a5c5873ef20870e1939e73cb9a3c1aeba5e3e72172a85ce9ed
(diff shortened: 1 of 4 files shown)


jjhelmus:
> validating dist/cpython-3.14.4-riscv64-unknown-linux-gnu-lto-20260422T1228.tar.zst
Error: errors found
  error: python/install/bin/python3.14 loads illegal library libgcc_s.so.1
  error: python/install/lib/libpython3.14.so.1.0 loads illegal library libgcc_s.so.1
  error: python/install/lib/python3.14/lib-dynload/_ctypes_test.cpython-314-riscv64-linux-gnu.so loads illegal library libgcc_s.so.1
  
  `libgcc_s.so.1` can be added to the allowed libraries, similar to armv7 targets. I am interested in why this occurs with LLVM and only when LTO is enabled.

jjhelmus:
Locally `python` in including symbols that do come from libgcc_s:

```
# readelf -sW ./python_lto/install/bin/python | grep "GCC"
   596: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __extenddftf2@GCC_3.0 (9)
   597: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __trunctfdf2@GCC_3.0 (9)
 ```
 
These are coming from a few different places, Modules/mathmodule.o is one example. My guess is that without LTO these are inlined.

jjhelmus:
closing, this was reworked as an experiment in #1259

PR #1104, opened 2026-04-23
remove tcl/tk translations and demo files from distributions

### LINKED ISSUE #1095: Consider not shipping lib/tk9.0/demos to save 1 MB
I'm about 99% sure people don't need this, but I wonder if demos/images gets inadvisably used somewhere.

```
$ du -sh "$(dirname "$(uv python find)")"/../lib/tk9.0/demos
1.2M    /home/ubuntu/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/bin/../lib/tk9.0/demos
```

Also maybe worth deleting: Tcl translations (I assume we don't ship translation files for Python itself, right? do we?)

```
$ du -sh "$(dirname "$(uv python find)")"/../lib/tcl9.0/msgs
512K    /home/ubuntu/.local/share/uv/python/cpython-3.14-linux-x86_64-gnu/bin/../lib/tcl9.0/msgs
```

jjhelmus:
Remove the Tcl/Tk translation files (.msg) and the Tk demos from the distribution artifacts. The former would only be needed in edge cases and the later is never be needed.

closes #1095

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -1379,6 +1379,10 @@ fi
 # Prune the tk demos which are > 1 MB and not used
 rm -rf "${ROOT}/out/python/install/lib/tk9.0/demos"
 
+# Prune the tcl/tk translations
+rm -rf "${ROOT}/out/python/install/lib/tcl9.0/msgs"
+rm -rf "${ROOT}/out/python/install/lib/tk9.0/msgs"
+
 # Copy the terminfo database if present.
 if [ -d "${TOOLS_PATH}/deps/usr/share/terminfo" ]; then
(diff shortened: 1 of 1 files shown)


jjhelmus:
On macOS, CPython 3.14, and the debug build, removing these files saves around ~450 kB in the compressed distribution and ~1200 kB when unpacked.

```
❯ ls -l *.tar.zst
-rw-r--r--@ 1 jjhelmus  staff  30127303 Apr 23 10:55 ref.tar.zst
-rw-r--r--@ 1 jjhelmus  staff  29677725 Apr 23 10:52 smaller.tar.zst
❯ tar xf ref.tar.zst -C ./ref
❯ tar xf smaller.tar.zst -C ./smaller
❯ du -Aks ref
197143  ref
❯ du -Aks smaller
195958  smaller
```

zanieb (approved):

geofft (approved):

PR #1114, opened 2026-05-01
update patchelf 0.13.1 -> 0.18.0

jjhelmus:
Update patchelf from 0.13.1 to 0.18.0.
This may address issues with loongarch64, see #1106

Issue #1106, opened 2026-04-25
Inconsistent page-size on loongarch64


wojiushixiaobai:
I encountered a **Segmentation Fault** when running a Python binary built for `loongarch64-unknown-linux-gnu`. After investigation, the root cause is that the ELF segments were realigned/restructured using a page size (likely 4KB) that is incompatible with the target system's kernel page size (16KB or 64KB).

This is a **similar issue** to:
- https://github.com/pypa/manylinux/issues/735
- https://github.com/numpy/numpy/issues/16677

```sh
root@linux-loong64:/tmp/build# unzip cpython-3.10-loongarch64-unknown-linux-gnu-lto.zip 
Archive:  cpython-3.10-loongarch64-unknown-linux-gnu-lto.zip
 extracting: cpython-3.10.20-loongarch64-unknown-linux-gnu-lto-20260411T1217.tar.zst  
root@linux-loong64:/tmp/build# tar -xf cpython-3.10.20-loongarch64-unknown-linux-gnu-lto-20260411T1217.tar.zst 
root@linux-loong64:/tmp/build# cd python/install/bin/
root@linux-loong64:/tmp/build/python/install/bin# /opt/clang/bin/llvm-strip --strip-debug python -o python-strip
root@linux-loong64:/tmp/build/python/install/bin# ./python-strip -V
段错误                  ./python-strip -V
root@linux-loong64:/tmp/build/python/install/bin# /opt/clang/bin/readelf -l python python-strip 

python

Elf file type is DYN (Shared object file)
Entry point 0x151288
There are 13 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000000040 0x0000000000000040 0x0002d8 0x0002d8 R   0x8
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x149640 0x149640 R E 0x4000
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
  LOAD           0x150000 0x0000000000150000 0x0000000000150000 0xf4b1b8 0xf4b1b8 R E 0x10000
  GNU_EH_FRAME   0xef617c 0x0000000000ef617c 0x0000000000ef617c 0x030b24 0x030b24 R   0x4
[...]

PR #1115, opened 2026-05-01
update patchelf 0.13.1 -> 0.14.5

jjhelmus:
Update patchelf from 0.13.1 to 0.14.5.
This may address issues with loongarch64, see https://github.com/astral-sh/python-build-standalone/issues/1106

Starting with 0.15.x, a compiler that supports C++17 is required to build patchelf. This is not available when cross-compiling.
See #1114


jjhelmus:
0.14.5 (and other 0.14.x) do require C++17 support as they use std::optional. This check was not added to configure until 0.15.x.

PR #1116, opened 2026-05-01
cross compile riscv64 with LLVM toolchain and sysroot

jjhelmus:
Use the same LLVM toolchain that is used for native x86-64 builds to cross compile for riscv64 using a sysroot.

### DIFF
--- a/cpython-unix/build.cross-riscv64.Dockerfile
+++ b/cpython-unix/build.cross-riscv64.Dockerfile
@@ -29,11 +29,12 @@ RUN for s in debian_buster debian_buster-updates debian-security_buster/updates;
 
 RUN apt-get update
 
-# Host building.
+# Build tools, same as in build.Dockerfile
 RUN apt-get install \
     bzip2 \
-    gcc \
-    g++ \
+    ca-certificates \
+    curl \
+    file \
     libc6-dev \
     libffi-dev \
     make \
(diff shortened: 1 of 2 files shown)


jjhelmus:
This differs in the approach in #1101 in that the LLVM toolchain is coming from the "custom compiled toolchain" (the `llvm-x86_64-linux` download target) that is also used to build the x86-64 targets. #1101 uses the toolchain provided by the LLVM project itself. This approach is more reproducible as the toolchain is pinned to a specific version and source.

jjhelmus:
Only the `libc6-dev-riscv64-cross` and `libgcc-8-dev-riscv64-cross` packages are needed for the sysroot, the others are dependencies of these but are included for clarity.

EliteTK (approved):

zanieb (approved):

jjhelmus:
I was ran some benchmarks on a riscv64 machine using the latest release of pbs and a build with this change. Using the average of three runs of [fastmark](https://pypi.org/project/fastmark/) using the LTO build of 3.14.5:

Latest release, 20260510 (GCC 8.3.0) : 778
With change, 20260522 (Clang 22.1.3) : 849

This is a ~9% increase in performance. Seems worthwhile to adopt this approach to the other platforms which are cross compiled.

PR #1122, opened 2026-05-06
bundle fallback Root Certificates

ewdurbin:
this is a common foot-gun when trying to build minimal images using python build standalone.

rather than relying on the user 1) ensuring root certificates are added to the final image AND 2) those certificates are placed in the correc place: bundle the Mozilla curated list from certifi.

for users who _definitely_ do not want any root certificates added, provide an opt-out via PYTHON_BUILD_STANDALONE_DISABLE_CA_FALLBACK


jjhelmus:
Should this be limited to Linux? I don't believe the fallback would ever be needed on macOS.

ewdurbin:
> Should this be limited to Linux? I don't believe the fallback would ever be needed on macOS.

good point! addressed in [519e052](https://github.com/astral-sh/python-build-standalone/pull/1122/commits/519e052b260ea6de2efca23bb27be6e3568af020)

PR #1123, opened 2026-05-07
CPython 3.15.0a8 -> 3.15.0b1

jjhelmus:
* Update CPython 3.15 to 3.15.0b1


zanieb (approved):

jjhelmus:
We should not release b1 without a fix for https://github.com/python/cpython/issues/149504 as it will break core functionality of `uv`

jjhelmus:
The paths used during windows builds changed in https://github.com/python/cpython/pull/149218

PR #1127, opened 2026-05-12
Strip toolchain bin paths from sysconfig AR variable

kimjune01:
## Summary
- Strip the toolchain bin directory prefix from sysconfig values so AR (and similar tools like RANLIB) contain just the binary name instead of build-time absolute paths
- Handles resolved symlinks (e.g. macOS `/var` vs `/private/var`) as a second pass
- Adds test covering 8 edge cases including trailing slashes and double-slash paths

## Test plan
- `python3 test_ar_normalization.py`: all 8 test cases passed


kimjune01:
Test passes on master too; appears already fixed. Closing.

PR #1130, opened 2026-05-13
Remove site startup patch

zsol:
## Summary
Remove the CPython 3.15 `site.py` reentrant startup patch corresponding to https://github.com/python/cpython/pull/149516 - a better version (https://github.com/python/cpython/pull/149659) has been merged 

> [!NOTE]
> merge this along with the bump to 3.15b2


zanieb (approved):
Thanks!

jjhelmus:
The patch in question was removed in #1136

PR #1132, opened 2026-05-20
Build OpenSSL with no-uplink for Python 3.12+

### LINKED ISSUE #640: OPENSSL_Applink crash on Windows
Reported in Discord: https://discord.com/channels/1039017663004942429/1379198677322694666/1379198677322694666

Using the python-build-standalone release from uv 0.7.5, `import requests` crashes the interpreter with `OPENSSL_Uplink(00007FFD958A4AF8,08): no OPENSSL_Applink`.

I cannot reproduce this crash, with either that version or the latest. I'm not totally sure what's going on.

(We do patch the uplink discovery code a bit, but we don't log what we're doing. Maybe we should.)

jjhelmus:
Following the behavior of CPython which dropped uplink support in 3.12, build OpenSSL with the `no-uplink` option starting with CPython 3.12. 

This allows a SSLContext to be created even when a SSLKEYLOGFILE environment variable is set on Windows.

closes #640

### DIFF
--- a/cpython-windows/build.py
+++ b/cpython-windows/build.py
@@ -674,17 +674,17 @@ def hack_project_files(
         pass
 
     # Our custom OpenSSL build has applink.c in a different location from the
-    # binary OpenSSL distribution. This is no longer relevant for 3.12+ per
-    # https://github.com/python/cpython/pull/131839, so we allow it to fail.swe
-    try:
+    # binary OpenSSL distribution.
+    # Starting with 3.12 CPython on Windows is built without uplink support
+    # so this modification is not needed.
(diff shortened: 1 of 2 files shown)


jjhelmus:
Upstream uplink support was removed in https://github.com/python/cpython/pull/131839. This was backported to 3.12 and 3.13 but not 3.11 or 3.10.

jjhelmus:
In addition to the disttests, I have build and tested Python 3.14 and 3.10 build using this change on a local Windows machine. An SSLContext as well as https requests can be made with or without a SSLKEYLOGFILE environment variable with this change.

zanieb (approved):

PR #1136, opened 2026-06-02
Bump 3.15.0 version

jjhelmus:
3.15.0b1 -> 3.15.0b2.

Remove patch introduced in #1123 as a different fix was merged upstream in python/cpython#149799
This makes #1130 unnecessary


zanieb (approved):

jjhelmus:
The upstream 3.15 build on Windows now uses Tcl/Tk 9.0.3 and sqlite 3.53.1. Updates are needed for these.

cpython-windows/build.py
```diff
@@ -1395,15 +1373,16 @@ def build_cpython(
...
-    # version there too
-    tk_bin_entry = (
-        "tk-windows-bin"
-        if meets_python_minimum_version(python_version, "3.14") or arch == "arm64"
-        else "tk-windows-bin-8612"
-    )
+    # We match the upstream tcl/tk version when possible
+    # 3.15 : 9.0.3
+    # 3.14 and arm64 : 8.6.14
+    # Others : 8.6.12
```
jjhelmus: This comment is not correct. I have a fix and will push if/when CI completes.

PR #1137, opened 2026-06-02
sqlite 3.50.4 -> 3.53.1

jjhelmus:


jjhelmus:
This is not necessary for the 3.15.0b2 release but is included in the Windows build upstream. It is worth including with or without the CPython version bump. The two changes are not coupled.

PR #1138, opened 2026-06-02
Remove unused entries in downloads module.

jjhelmus:


geofft (approved):

geofft:
`tix` needs to stay because it's still used on Windows on older Python versions and still works fine. I think this could be merged if you drop that commit and remove the other two entries?

jjhelmus:
Right, `tix` was not removed from CPython until 3.13. From [What’s New In Python 3.13](https://docs.python.org/3.13/whatsnew/3.13.html):
> Remove the tkinter.tix module (deprecated in Python 3.6).

That said, the `tix` download entry is not used. Tcl/Tk on Windows is provided by a prebuilt binary from [python/cpython-bin-deps](https://github.com/python/cpython-bin-deps) which includes tix in the tcltktk-8.6.* tags. The `tix` entry refers to the source which is not used during the build. It does serve to pull in the `LICENSE.tix.txt` file.

PR #1145, opened 2026-06-10
Bump stable Python versions: 3.13.14 and 3.14.6

zanieb:
* Update CPython 3.13.13 -> 3.13.14
* Update CPython 3.14.5 -> 3.14.6


geofft:
I think this is the same as #1146 but without one necessary fix?

zanieb:
Sure. It came first, I just kicked it off in the background.

PR #1147, opened 2026-06-10
Cherry-pick an improved DER parser in _ssl.c from 3.12 onto 3.10 and 3.11

geofft:
This fixes the inability to load DER correctly with OpenSSL 3.5.7.


jjhelmus:
Should this patch be applied on Windows, e.g. in `cpython-windows/build.py`?

geofft:
That seems like a really good idea, yes...

I'm including the patch twice, once in cpython-unix and once in cpython-windows, to avoid thinking about symlinks or about .., but I can read the patch out of cpython-unix if that seems better.

jjhelmus (approved):

jjhelmus:
Thanks!

PR #1148, opened 2026-06-10
Use v4 DWARF on riscv64

jjhelmus:
Use DWARF v4 on riscv64 to avoid relocations that the GNU linker from Debian Buster does not support.

### DIFF
--- a/cpython-unix/targets.yml
+++ b/cpython-unix/targets.yml
@@ -451,6 +451,9 @@ riscv64-unknown-linux-gnu:
   host_cxx: clang++
   target_cc: /usr/bin/riscv64-linux-gnu-clang
   target_cxx: /usr/bin/riscv64-linux-gnu-clang++
+  target_cflags:
+    # Keep debug relocations old enough for Buster's riscv64 GNU ld.
+    - '-fdebug-default-version=4'
   target_ldflags:
     # Hardening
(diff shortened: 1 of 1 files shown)


zanieb (approved):

jjhelmus:
This addresses the unsupported relocation types when linking that have existed since the switch to LLVM with riscv in #1101. 

Specifically these types of warnings are no longer present as DWARF v4 does not uses these types of relocations.
```
cpython-3.13> /usr/bin/riscv64-linux-gnu-ld: Objects/listobject.o: unsupported relocation type 0x3c
cpython-3.13> /usr/bin/riscv64-linux-gnu-ld: Objects/listobject.o: unsupported relocation type 0x3d
```

Unfortunately this does not address the build failure on 3.13 freethreaded+debug that started with the update to OpenSSL 3.5.7, #1144.
```
2026-06-10T22:07:32.1897413Z cpython-3.13> ciphercommon.c:(.text+0x1534): relocation truncated to fit: R_RISCV_JAL against symbol `ERR_set_error' defined in .text section in /tools/deps/lib/libcrypto.a(libcrypto-lib-err_blocks.o)
```

jjhelmus:
This is no longer needed now that GCC is used for riscv64 cross compiling, #1150. This option should be revisited if LLVM is once again used.

PR #1150, opened 2026-06-10
Revert "cross compile riscv64 with LLVM toolchain and sysroot (#1116)"

jjhelmus:
This reverts commit 96a589a4042eaa37acb695cd70bd4ac925803994.

Having a split LLVM / older GNU ld toolchain causes issues with relocation symbols and truncations.

### DIFF
--- a/cpython-unix/build.cross-riscv64.Dockerfile
+++ b/cpython-unix/build.cross-riscv64.Dockerfile
@@ -29,12 +29,11 @@ RUN for s in debian_buster debian_buster-updates debian-security_buster/updates;
 
 RUN apt-get update
 
-# Build tools, same as in build.Dockerfile
+# Host building.
 RUN apt-get install \
     bzip2 \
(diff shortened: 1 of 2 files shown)


zanie-oai:
Why are you running things like this with all Python versions and build variants? jw, seems unnecessary and want to understand if the subsetting is working

zanieb:
Merging assuming this is blocking release

jjhelmus:
> Why are you running things like this with all Python versions and build variants?

`main` was failing on a single build of riscv64 due to relocation relaxation. The specifics of this were highly dependent on minor changes, 3.13 was failing before the upgrade to 3.13.14. 3.14 failed with the update to 3.14.6, the DER parser modification caused more builds to fail.

I wanted to ensure that reverting to GCC did not introduce new failures and fixed and fixed the current ones. The best method seemed to be to test everything that could be effected.

PR #1152, opened 2026-06-12
safely remove PYTHON* env variables in distribution tests

jjhelmus:
Safely remove PYTHON* environment variables when running the distibution tests.

Since the env dictionary is mutated within the loop a copy must be made.


jjhelmus:
This is not causing an issue in CI as there are no `PYTHON*` environment variables set. I ran into this when testing locally.

Issue #1140, opened 2026-06-03
tkinter broken on 3.15.0b2 release on Windows


jjhelmus:
The latest 3.15.0b2 release (20260602) on Windows is missing the tcl library. This results in a broken `tkinter` module.

The layout script is unable to find the correct location during the build as can be seen in the log:
```
2026-06-03T01:36:34.0101544Z cpython> WARNING  Failed to find TCL_LIBRARY
```

jjhelmus:
PC/layout is being used to copy over the files from the Tcl/Tk archive into the install directory, specifically [this block](https://github.com/python/cpython/blob/v3.15.0b2/PC/layout/main.py#L110-L126). 

Debugging this locally I found that the `TCL_LIBRARY.env` file does exist and contains a path to a tcl9.0 directory that does not exist in the prebuilt tcl_tk artifact. Tcl 9 includes zipfs support which allows reading directly from zip files. The prebuild tcl_tk artifact packages the tcl and tk scripts in zip files but it is unclear where these should be placed in the install artifact. The layout script does not copy these zip files.

Upstream includes these files in a "tcl" directory but also has a broken tkinter in 3.15.0b2, https://github.com/python/cpython/issues/150836

jjhelmus:
To address this two change are needed:
* The fix from python/cpython#151562 which mounts the embedded Tk DLL when tkinit is initiated. This will be included in 3.15.0b3 scheduled for release tomorrow.
* The Tcl/Tk 9 DLLs need to retain their embedded zip files. Currently these are stripped. This is in progress in #1161

jjhelmus:
With the two changes outlines in the previous comment merged the 20260623 release of CPython 3.15.0b3 has a working tkinter.

I validated this on a Windows machine using the install_only_stripped artifact and `import tkinter; tkinter.Tk()` works as expected.

PR #1161, opened 2026-06-22
Retain embedded ZipFS data in Tcl/Tk DLLs on Windows

jjhelmus:
Do not strip the tcl90.dll and tcl9tk90.dll DLLs in the Windows distribution (which are now allowed) as they contained embedded versions of the Tcl and Tk library as ZipFS data that are required at runtime for Tkinter to work correctly.

Currently this is limited to CPython 3.15 as this is the only version that uses Tcl/Tk 9.0 on Windows. This may change if support for Tcl/Tk 9.0 is backported. 

Also update the Windows GitHub action to eheck the error code after the validate-distribution is run and end early if needed.

See https://github.com/actions/runner-images/issues/6668

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -186,4 +186,7 @@ jobs:
         run: |
           $Dists = Resolve-Path -Path "dist/*.tar.zst" -Relative
           .\pythonbuild.exe validate-distribution $Dists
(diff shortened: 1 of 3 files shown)


jjhelmus:
This PR will eventually provide part of the necessary fix for #1140 but first validate-distribution needs to fail due to the new Tcl/Tk 9 libraries.

PR #1162, opened 2026-06-23
CPython 3.15.0b3

jjhelmus:
CPython 3.15.0b2 -> 3.15.0b3


jjhelmus:
weakref.c was added to the test{limited}capi module. I have a fix but am waiting for the Windows builds to finish so that platform can be removed from the CI matrix.

Issue #193, opened 2023-10-08
Kernel-dependent features like `os.pidfd_open()`


achimnol:
Recent Python versions add optional features such as [`os.pidfd_open()`](https://docs.python.org/3/library/os.html#os.pidfd_open) when it is available during the build time.

It seems that the current Python 3.11 distribution in this repo does not have `os.pidfd_open()`.
While this is not a critical regression because most libraries depending on it has a good fallback, but I'd like to be able to adopt such new kernel features.

I'm not sure how the current release process could handle. Maybe we need to have multiple different builds like the x86_64 v2/v3/... CPU generations, by the Linux kernel version. I'm afraid that this would incur too much burden for the build infra.

I'm just reporting this issue as a future reference, though.

indygreg:
The way things currently work is that we build the CPython distributions against very old kernel headers and glibc to ensure maximum binary portability by targeting a very old, ~universally supported syscall + glibc API surface.

CPython currently uses compile time checks for features like `pidfd_open()` support. Literally a `#if defined(__linux__) && defined(__NR_pidfd_open)` in C source code. So if the feature isn't present at compile time, you don't get it at run time.

On macOS, CPython is using Mach-O weak symbols and run-time checks allow us to reference macOS SDK APIs that aren't available in all machines. If they are present at run-time, CPython can use them. If they aren't, the features depending on them aren't available. This is ideal from an end-user perspective because it doesn't penalize the common user running on modern macOS by depriving them of newer features.

On Linux, ELF has support for weak symbols/linking. This is conceptually similar to Mach-O's similar feature. Usually you add a compiler #pragma or similar preprocessor directive to indicate a symbol is weakly linked. If the symbol resolves at run-time, the symbol/function address is non-0 and you can call it.

Unfortunately, I don't believe CPython has any support for weak symbols on Linux/ELF. So getting run-time conditional features isn't trivially achievable. (I'd have to page this in my brain but I want to say there are some practical limitations of weak symbols on ELF that may make their use non-viable. Even if there are, there are similar features like IFUNC that could potentially be employed for dynamic run-time dispatch support.)

That's a long way of saying that features like pidfd_open() currently require separate build variants to work. That leads to an explosion of build variants targeting various Linux + glibc feature levels. That becomes unwieldy very fast.

[...]

PR #1163, opened 2026-06-24
Build CPython against modern kernel UAPI headers on *-linux-gnu targets

jjhelmus:
Use modern Linux kernel UAPI headers when building CPython for glibc Linux targets.

Adds a `linux-uapi` package to the project that contains the unpacked `linux-libc-dev` package from a modern Debian release, currently `trixie-backports`. This provides Linux UAPI headers from a modern kernel release.

Building CPython against these headers makes newer syscall definitions and macros available at build time, enabling CPython functionality such as `os.pidfd_open`.

This does not increase the minimum required glibc or runtime kernel version.

APIs compiled using these definitions may still fail at runtime. For example, if an API uses a direct syscall and the running kernel does not implement it, the kernel returns `ENOSYS`, which Python exposes as an `OSError`.

Consequently, functions such as `os.pidfd_open` will exist when running on an older kernel that does not implement the corresponding syscall. Applications should handle `OSError` with `errno.ENOSYS` when runtime kernel support is uncertain.

This should be included as a quirk of `python-build-standalone` and documented as such. The trade-off here is worth introducing this quirk.

Related to #193
Related to astral-sh/uv#11811, astral-sh/uv#9374

### DIFF
--- a/cpython-unix/Makefile
+++ b/cpython-unix/Makefile
@@ -220,6 +220,9 @@ $(OUTDIR)/xcb-proto-$(XCB_PROTO_VERSION)-$(PACKAGE_SUFFIX).tar: $(PYTHON_DEP_DEP
 $(OUTDIR)/xorgproto-$(XORGPROTO_VERSION)-$(PACKAGE_SUFFIX).tar: $(PYTHON_DEP_DEPENDS) $(HERE)/build-xorgproto.sh
 	$(RUN_BUILD) --docker-image $(DOCKER_IMAGE_BUILD) xorgproto
 
+$(OUTDIR)/linux-uapi-$(LINUX_UAPI_VERSION)-$(PACKAGE_SUFFIX).tar: $(PYTHON_DEP_DEPENDS) $(HERE)/build-linux-uapi.sh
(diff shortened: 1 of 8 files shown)


jjhelmus:
The include guard in posixmodule.py for sys/syscall.h is `HAVE_GETRANDOM_SYSCALL` which is falsely in glibc 2.19 which is used on x86-64. A better guard would be `HAVE_SYS_SYSCALL_H`.

geofft:
Oh, this is an interesting approach. I think it seems reasonable and safe!

Note that the quirk is technically already true of "normal" Python distributions if you run a newer OS image in e.g. a Docker container on a host with an older kernel. So generic code (libraries, etc.) already need to both test `hasattr` and catch `ENOSYS`, though very specific applications that are never deployed in containers may not have had to worry about this. I agree we should document it, but we should document it in a way that conveys it's a best practice anyway. (I think we also maybe should try to convince upstream CPython that, because of this, it would be nicer to have the functions be unconditionally present and fail with a synthetic `ENOSYS` if we don't have the right constants at compile time.)

zanieb (approved):

PR #1164, opened 2026-06-24
Enable os.memfd_create through runtime weak linking

jjhelmus:
Always build CPython with `os.memfd_create` but remove the Python function from the namespace at runtime when glibc does not provide the an implementation.

This is accomplished using a weak `memfd_create` symbol which is evaluated at runtime.

This requires the modern kernel UAPI headers from #1163 to define the `MFD_*` constants.


zanieb:
We should definitely try to upstream any weak linking implementations

zanieb (approved):

jjhelmus:
Rebased on top of merged #1163

PR #1165, opened 2026-06-25
run build.py in frozen mode

jjhelmus:
Run build.py with the `--frozen` option to prevent updates to `uv.lock`.


zanieb:
Why? I'd expect `--locked` if anything.

PR #1174, opened 2026-07-09
Validate that distributions include stack protection and fortify sources

### LINKED ISSUE #1172: Validate that distributions include stack protection and fortify sources
The post build validation should check that the binary has stack-protection features and uses fortify sources when applicable.

This is a follow up to #1166 and related to #837

jjhelmus:
Similar validation is done in the Debian `hardening-check` utility. 

closes #1172

### DIFF
--- a/src/validation.rs
+++ b/src/validation.rs
@@ -926,6 +926,18 @@ fn validate_elf<Elf: FileHeader<Endian = Endianness>>(
     elf: &Elf,
     data: &[u8],
 ) -> Result<()> {
+    // The target hardening flags are intended to protect the interpreter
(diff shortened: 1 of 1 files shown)


src/validation.rs
```diff
@@ -1120,6 +1132,17 @@ fn validate_elf<Elf: FileHeader<Endian = Endianness>>(
 
             for (symbol_index, symbol) in symbols.enumerate() {
                 let name = String::from_utf8_lossy(symbol.name(endian, strings)?);
+                let is_undefined = symbol.is_undefined(endian);
```
EliteTK: This variable name is a bit vague out of context.

EliteTK (approved):

Issue #227, opened 2024-03-10
Restore shared extension module and/or dependency libraries


indygreg:
This project was created in support of PyOxidizer. And a goal of PyOxidizer was to see if it was possible to achieve a single file executable Python application.

In order to achieve that goal, this project built highly statically linked Python distributions.

On Linux and macOS, extension modules are statically linked into libpython. And 3rd party library dependencies of extension modules (OpenSSL, SQLite, etc) are also statically linked into libpython.

This mostly _just works_. Until it doesn't.

For example, ELF symbols resolution prefers symbols in the global namespace over a local namespace. So if you load a .so providing symbol `foo` which is also provided by an external `libpython.so`, the `libpython.so` symbol is used. This leads to wonkiness like https://gregoryszorc.com/docs/python-build-standalone/main/quirks.html#static-linking-of-libx11-incompatibility-with-pyqt-on-linux.

As the Python distributions in this project have grown a life outside of PyOxidizer, the opinionated choice of statically linking all extension modules and shared libraries is hard to justify. We're foregoing compatibility with aspects of the larger Python ecosystem.

So I'm proposing that we revert to distributing 3rd party libraries and extension modules as shared libraries.

indygreg:
I looked at this over the weekend. I was soft blocked after running into a bug with `patchelf` corrupting ELF binaries. I need to read up on how the GNU dynamic loader handles rpath and `$ORIGIN` before I continue with an alternate solution.

mzpqnxow:
Just stumbled across this project and wish I had found it a year ago

I have a (much, much) less ambitious "project" (a couple of scripts really) [here](https://github.com/mzpqnxow/python-rpath-build) that may be helpful as a reference for `$ORIGIN` linking

My use-case is probably unusual -  I need to have Python linked against a few older as well as custom-patched OpenSSL versions. I didn't want to deal with setting runtime linking environment variables (e.g. LD_LIBRARY_PATH) and didn't want to pollute my system which is why I ended up with `$ORIGIN`. It's not exactly what this project aims to help with, but there's overlap in the requirements, obviously

The junk I have at that repo is not something that you'll be able to copy/paste but I hope it might be a little helpful - even if just to see what flags to use and how the syntax looks - because there's a `$` in `$ORIGIN`, and because the string is actually passed as a command there's some wacky double-escaping

[...]

PR #1180, opened 2026-07-17
Use standard extension configuration for CPython 3.12+

### LINKED ISSUE #645: Address HACL SIMD compilation in a better way
HACL has some optimized SIMD implementations that are conditionally compiled in by CPython's build system. Because we rework the build process for extension modules, we ended up with a build failing because these implementations were referenced but not compiled. The simplest thing to do for #610 was to change the build system to unconditionally choose not to use the SIMD implementations (patch-python-configure-hacl-no-simd.patch).

Also, I'm a little unsure whether this properly handles the case where the target machine doesn't actually have SIMD (x86_64-v2 for SIMD128, x86_64-v3 for SIMD256), so it seemed safer to turn it off until confirming that. Also, there are some comments about the fact that the performance is worse on arm64.

We should investigate this in more detail. I imagine that we'll probably want to fix (or ideally get rid of) our custom build for extension modules. We may also need to submit something to upstream CPython to properly handle the case where we're building with a toolchain that supports SSE/AVX but we want to produce binaries that work without it.

For reference, here's the relevant code in configure.ac (as of HEAD right now, i.e. 3.15.0a):
```autoconf
dnl Check if universal2 HACL* implementation should be used.
if test "$UNIVERSAL_ARCHS" = "universal2" -o \
   \( "$build_cpu" = "aarch64" -a "$build_vendor" = "apple" \)
then
  use_hacl_universal2_impl=yes
else
  use_hacl_universal2_impl=no
fi

# The SIMD files use aligned_alloc, which is not available on older versions of
# Android.
# The *mmintrin.h headers are x86-family-specific, so can't be used on WASI.
if test "$ac_sys_system" != "Linux-android" -a "$ac_sys_system" != "WASI" || \
   { test -n "$ANDROID_API_LEVEL" && test "$ANDROID_API_LEVEL" -ge 28; }
then
  dnl This can be extended here to detect e.g. Power8, which HACL* should also support.
  AX_CHECK_COMPILE_FLAG([-msse -msse2 -msse3 -msse4.1 -msse4.2],[
    [LIBHACL_SIMD128_FLAGS="-msse -msse2 -msse3 -msse4.1 -msse4.2"]

    AC_DEFINE([HACL_CAN_COMPILE_SIMD128], [1], [HACL* library can compile SIMD128 implementations])

    # macOS universal2 builds *support* the -msse etc flags because they're
    # available on x86_64. However, performance of the HACL SIMD128 implementation
    # isn't great, so it's disabled on ARM64.
    AC_MSG_CHECKING([for HACL* SIMD128 implementation])
    if test "$use_hacl_universal2_impl" = "yes"; then
      [LIBHACL_BLAKE2_SIMD128_OBJS="Modules/_hacl/Hacl_Hash_Blake2s_Simd128_universal2.o"]
      AC_MSG_RESULT([universal2])
    else
      [LIBHACL_BLAKE2_SIMD128_OBJS="Modules/_hacl/Hacl_Hash_Blake2s_Simd128.o"]
      AC_MSG_RESULT([standard])
    fi

  ], [], [-Werror])
fi
AC_SUBST([LIBHACL_SIMD128_FLAGS])
AC_SUBST([LIBHACL_BLAKE2_SIMD128_OBJS])

# The SIMD files use aligned_alloc, which is not available on older versions of
# Android.
# The *mmintrin.h headers are x86-family-specific, so can't be used on WASI.
#
# Although AVX support is not guaranteed on Android
# (https://developer.android.com/ndk/guides/abis#86-64), this is safe because we do a
# runtime CPUID check.
if test "$ac_sys_system" != "Linux-android" -a "$ac_sys_system" != "WASI" || \
   { test -n "$ANDROID_API_LEVEL" && test "$ANDROID_API_LEVEL" -ge 28; }
then
  AX_CHECK_COMPILE_FLAG([-mavx2],[
    [LIBHACL_SIMD256_FLAGS="-mavx2"]
    AC_DEFINE([HACL_CAN_COMPILE_SIMD256], [1], [HACL* library can compile SIMD256 implementations])

    # macOS universal2 builds *support* the -mavx2 compiler flag because it's
    # available on x86_64; but the HACL SIMD256 build then fails because the
    # implementation requires symbols that aren't available on ARM64. Use a
    # wrapped implementation if we're building for universal2.
    AC_MSG_CHECKING([for HACL* SIMD256 implementation])
    if test "$use_hacl_universal2_impl" = "yes"; then
      [LIBHACL_BLAKE2_SIMD256_OBJS="Modules/_hacl/Hacl_Hash_Blake2b_Simd256_universal2.o"]
      AC_MSG_RESULT([universal2])
    else
      [LIBHACL_BLAKE2_SIMD256_OBJS="Modules/_hacl/Hacl_Hash_Blake2b_Simd256.o"]
      AC_MSG_RESULT([standard])
    fi
  ], [], [-Werror])
fi
AC_SUBST([LIBHACL_SIMD256_FLAGS])
AC_SUBST([LIBHACL_BLAKE2_SIMD256_OBJS])
### end(HACL* configuration)
```

and Makefile.pre.in does
```make
LIBHACL_BLAKE2_SIMD128_CFLAGS=@LIBHACL_SIMD128_FLAGS@ -DHACL_CAN_COMPILE_VEC128
LIBHACL_BLAKE2_SIMD256_CFLAGS=@LIBHACL_SIMD256_FLAGS@ -DHACL_CAN_COMPILE_VEC256
```
```make
LIBHACL_BLAKE2_SIMD128_OBJS=@LIBHACL_BLAKE2_SIMD128_OBJS@
LIBHACL_BLAKE2_SIMD256_OBJS=@LIBHACL_BLAKE2_SIMD256_OBJS@
LIBHACL_BLAKE2_OBJS= \
                Modules/_hacl/Hacl_Hash_Blake2s.o \
                Modules/_hacl/Hacl_Hash_Blake2b.o \
                Modules/_hacl/Lib_Memzero0.o \
                $(LIBHACL_BLAKE2_SIMD128_OBJS) \
                $(LIBHACL_BLAKE2_SIMD256_OBJS)
```
```make
Modules/_hacl/Hacl_Hash_Blake2s.o: $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2s.c $(LIBHACL_BLAKE2_HEADERS)
        $(CC) -c $(LIBHACL_CFLAGS) -o $@ $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2s.c
Modules/_hacl/Hacl_Hash_Blake2b.o: $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2b.c $(LIBHACL_BLAKE2_HEADERS)
        $(CC) -c $(LIBHACL_CFLAGS) -o $@ $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2b.c
Modules/_hacl/Hacl_Hash_Blake2s_Simd128.o: $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2s_Simd128.c $(LIBHACL_BLAKE2_HEADERS)
        $(CC) -c $(LIBHACL_CFLAGS) $(LIBHACL_BLAKE2_SIMD128_CFLAGS) -o $@ $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2s_Simd128.c
Modules/_hacl/Hacl_Hash_Blake2s_Simd128_universal2.o: $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2s_Simd128_universal2.c $(LIBHACL_BLAKE2_HEADERS)
        $(CC) -c $(LIBHACL_CFLAGS) $(LIBHACL_BLAKE2_SIMD128_CFLAGS) -o $@ $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2s_Simd128_universal2.c
Modules/_hacl/Hacl_Hash_Blake2b_Simd256.o: $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2b_Simd256.c $(LIBHACL_BLAKE2_HEADERS)
        $(CC) -c $(LIBHACL_CFLAGS) $(LIBHACL_BLAKE2_SIMD256_CFLAGS) -o $@ $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2b_Simd256.c
Modules/_hacl/Hacl_Hash_Blake2b_Simd256_universal2.o: $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2b_Simd256_universal2.c $(LIBHACL_BLAKE2_HEADERS)
        $(CC) -c $(LIBHACL_CFLAGS) $(LIBHACL_BLAKE2_SIMD256_CFLAGS) -o $@ $(srcdir)/Modules/_hacl/Hacl_Hash_Blake2b_Simd256_universal2.c
```

### LINKED ISSUE #869: Upstream patch `patch-checksharedmods-disable.patch`
[`patch-checksharedmods-disable.patch`](https://github.com/astral-sh/python-build-standalone/blob/main/cpython-unix/patch-checksharedmods-disable.patch)

There's a post-build Python script that verifies modules were built correctly. Our custom Setup.local-based module building completely breaks assumptions in this script.

This patch disables the verification script. Ideally we'd fix the script instead.

### LINKED ISSUE #875: Upstream patch `patch-configure-disable-stdlib-mod-3.12.patch`
[`patch-configure-disable-stdlib-mod-3.12.patch`](https://github.com/astral-sh/python-build-standalone/blob/main/cpython-unix/patch-configure-disable-stdlib-mod-3.12.patch)

Python 3.11+ has configure support for configuring extension modules. However, the functionality doesn't support all extension modules and doesn't easily support static linking of extra libraries (a limitation of `makesetup`).

This patch disables the configure-based module configuration and uses our auto-generated Setup.local instead.

**Related:**
- [`patch-configure-disable-stdlib-mod.patch`](https://github.com/astral-sh/python-build-standalone/blob/main/cpython-unix/patch-configure-disable-stdlib-mod.patch) (for Python 3.11)

### LINKED ISSUE #876: Upstream patch `patch-pwd-remove-conditional.patch`
[`patch-pwd-remove-conditional.patch`](https://github.com/astral-sh/python-build-standalone/blob/main/cpython-unix/patch-pwd-remove-conditional.patch)

Disabling the configure-based module configuration (see patch-configure-disable-stdlib-mod-3.12.patch) breaks the conditional definition of the pwd module in Setup.bootstrap.in.

This patch removes that conditional to work around the issue.

### LINKED ISSUE #883: Upstream patch `patch-python-configure-hacl-no-simd.patch`
[`patch-python-configure-hacl-no-simd.patch`](https://github.com/astral-sh/python-build-standalone/blob/main/cpython-unix/patch-python-configure-hacl-no-simd.patch)

The configure script tries to auto-detect whether it can build 128-bit and 256-bit SIMD helpers for HACL. On x86-64 that requires v2 and v3 microarchitecture levels respectively, and on arm64 the performance is poor.

This patch disables SIMD optimization auto-detection for HACL.

jjhelmus:
Switch CPython 3.12+ builds to the upstream Setup.stdlib extension-building mechanism while preserving the existing static/shared linkage behavior.

* Build pkgconf on macOS so that bundled dependencies can be discovered and configured. 
* Generate a minimal Setup.local for shared and disabled overrides on CPython 3.12+.
* Restore CPython’s post-build extension checks and remove obsolete patches/workarounds.
* Preserve SQLite extension loading and macOS symbol hiding.
* Explicitly disable the clinic test and `xxlimited_3_13` extensions. c.f. #1185 

With this change, `extension-modules.yml` is still used for the following in CPython 3.12:
* Disabling modules
* Specifying static vs shared linkage
* Validating the extension modules specified in the CPython source archive.
* Constructing `PYTHON.json`

Constructing `PYTHON.json` for CPython 3.12+ from other sources is a needed follow up.

Makes various patches unnecessary:
* closes #869 
* closes #875 (This patch is only applied on Python 3.11)
* closes #876 (this patch is only applied on Python 3.11)

Configuration can now appropriately detect SIMD support in the toolchain and adjusts `libhacl` flags accordingly.
* closes #883
* closes #645
  
This does not add shared extension suggested in #227. This could be done with a small change as a follow up.


src/validation.rs
```diff
@@ -1826,12 +1826,24 @@ fn validate_extension_modules(
...
         if !static_crt {
             wanted.insert("_testcapi");
 
             if !matches!(python_major_minor, "3.10" | "3.11" | "3.12") {
                 wanted.insert("_testlimitedcapi");
             }
+
+            if python_major_minor == "3.15" {
+                wanted.insert("xxlimited_3_13");
+            }
```
zanieb: Why these changes?

zanieb: > Enable the clinic test extensions and handle xxlimited_3_13 appropriately.

This doesn't really seem like it belongs here?

It seems like we should make that decision independently and this should be a pure refactor...

jjhelmus: These modules need to be declared in `extension-modules.yaml` as this change now checks for missing extensions modules against `Setup.stdlib.in`. The existing behavior of not building them can be maintained and the decision to enable these extensions done as a follow up.

I'll update and create an issue to track the decision on if these extensions should be included.

jjhelmus: c.f. #1185

jjhelmus:
With `MODULE_BUILDTYPE=static` set CPython 3.12+ still builds a handful of extensions modules as shared, specifically: 
* `_ctypes_test`
* `_testimportmultiple`
* `_testmultiphase`
* `_testsinglephase`
* `xxlimited`
* `xxlimited_35`
* `xxlimited_3_13`

Using the custom configuration the last three of these are disabled and not built. The first four are build as shared modules except for static builds where they are statically linked. Because of the behavior of Setup.stdlib.in theses three need to be included in Setup.local to restore the static linkage. This should be revisited as disabling these in the static builds may be more prudent.

The existing behavior for 3.10 and 3.11 is to include a shared module, #1222 which is retained.
No shared modules are included in 3.12+ which is replicated here and validated.

jjhelmus:
This is ready for a review. 

There is follow up work to construct `PYTHON.json` from source artifacts rather than from `extension-modules.yml` for 3.12+ as the yaml file is no longer the source of truth. 

This could be done in this PR, in a stacked PR, or as a fast follow. I'd prefer merging this and following with another PR to avoid a long running branch. The work for building `PYTHON.json` without the yaml file is straightforward but benefits from refactoring the long `derive_setup_local` function for clarity.

cpython-unix/build-cpython.sh
```diff
@@ -641,13 +643,6 @@ if [[ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_13}" && "${TARGET_TRIPLE}" == x86_64*
     PROFILE_TASK="${PROFILE_TASK} --ignore test.test_bytes.BytesTest.test_from_format"
 fi
 
-# ./configure tries to auto-detect whether it can build 128-bit and 256-bit SIMD helpers for HACL,
-# but on x86-64 that requires v2 and v3 respectively, and on arm64 the performance is bad as noted
-# in the comments, so just don't even try. (We should check if we can make this conditional)
-if [[ -n "${PYTHON_MEETS_MINIMUM_VERSION_3_14}" ]]; then
-    patch -p1 -i "${ROOT}/patch-python-configure-hacl-no-simd.patch"
-fi
```
zanieb, line 649: Why is this resolved?

cpython-unix/build-cpython.sh
```diff
@@ -1112,9 +1113,6 @@ if linux_uapi_include_arch:
         "",
     )
 replace_in_all("-isystem %s/deps/linux-uapi/usr/include" % tools_path, "")
-# See https://github.com/python/cpython/issues/145810#issuecomment-4068139183
-replace_in_all("-LModules/_hacl", "")
```
zanieb, line 1116: Same as https://github.com/astral-sh/python-build-standalone/pull/1180/changes#r3814055392

cpython-unix/build-tcl.sh
```diff
@@ -83,6 +83,12 @@ export CPPFLAGS="${CFLAGS}"
 # Initialize it so that package configure does not require an installed Tcl 9.
 export ZIPFS_BUILD=0
 
+# CPython uses the macOS SDK's zlib without a zlib.pc file. Keep the -lz in
+# Libs.private, but do not make pkg-config require the missing metadata.
+if [[ "${PYBUILD_PLATFORM}" = macos* ]]; then
+    sed -i '' -e 's/ zlib >= 1.2.3//' tcl.pc.in
```
zanieb: Should we be using a patch here instead?

cpython-unix/build-tk.sh
```diff
@@ -31,6 +31,10 @@ if [[ "${PYBUILD_PLATFORM}" = macos* ]]; then
 else
     LDFLAGS="${LDFLAGS} -Wl,--exclude-libs,ALL"
     EXTRA_CONFIGURE_FLAGS="--x-includes=${TOOLS_PATH}/deps/include --x-libraries=${TOOLS_PATH}/deps/lib"
+
+    # Make pkg-config --static resolve libX11's private xcb and Xau dependencies.
+    sed -i '/^Requires: /a\
+Requires.private: x11' tk.pc.in
```
zanieb: Same as https://github.com/astral-sh/python-build-standalone/pull/1180/changes#r3814061876

cpython-unix/build-cpython.sh
jjhelmus: Portions of this comment are incorrect, the behavior of configure is safe.

The check that configure performs is to determine if the compiler supports the necessary flags (`-msse`, `-mavx2`, etc) to allow the SIMD helpers to be built. 

On Linux x86-64 these flags are supported by LLVM so the helpers are built:
```
cpython-3.14> checking whether C compiler accepts -msse -msse2 -msse3 -msse4.1 -msse4.2...
cpython-3.14> yes
cpython-3.14> checking for HACL* SIMD128 implementation...
cpython-3.14> standard
cpython-3.14> checking whether C compiler accepts -mavx2...
cpython-3.14> yes
cpython-3.14> checking for HACL* SIMD256 implementation...
cpython-3.14> standard
```

A [runtime check](https://github.com/python/cpython/blob/v3.14.7/Modules/blake2module.c#L331-L352) is performed before dispatching to these function, the availability of these helpers does not change the baseline x86-64 requirement.

On macOS arm64 the check fails and the SIMD helpers are not built:
```
cpython-3.14> checking whether C compiler accepts -msse -msse2 -msse3 -msse4.1 -msse4.2... no
cpython-3.14> checking whether C compiler accepts -mavx2... no
```

On macOS x86-64 the check succeeds and the universal2 versions of the SIMD helpers are built. The universal2 wrapper prevents the SIMD implementations from being included when building the arm64 portion of the universal2 library but this never comes into play because only the x86-64 portion of the library is built.

```
cpython-3.14> checking whether C compiler accepts -msse -msse2 -msse3 -msse4.1 -msse4.2... yes
cpython-3.14> checking for HACL* SIMD128 implementation... universal2
cpython-3.14> checking whether C compiler accepts -mavx2... yes
cpython-3.14> checking for HACL* SIMD256 implementation... universal2
```

The runtime check is done so these helpers are only used when the host indicates support. When running on native x86-64 hardware this depends on the host CPU. When running on Apple Silicon under Rosetta2, the behavior depends on what features Rosetta advertises. The performance is intended by upstream and can be optimized by the user by using to the build that matches the host architecture.

cpython-unix/build-cpython.sh
jjhelmus: This flag is no longer being added to LDFLAGS (see the removal of lines 388-394) so it no longer leaks into sysconfig.

jjhelmus: For verification I checked the `cpython-3.14-x86_64-unknown-linux-gnu-pgo+lto` artifact:

```
❯ unzip cpython-3.14-x86_64-unknown-linux-gnu-pgo+lto.zip
...
❯ tar xf cpython-3.14.7-x86_64-unknown-linux-gnu-pgo+lto-20260818T1407.tar.zst
❯ cd python/install
❯ rg LModules/_hacl

❯
```

The pattern that used to be replaced does not appear in sysconfig or elsewhere.

cpython-unix/build-tcl.sh
jjhelmus: This could be accomplished by a patch but I think that setting `TCLTK_CFLAGS` and `TCLTK_LIBS` when building CPython on macOS is cleaner. That approach makes this change unnecessary, matches how CPython's macOS installer builds `_tkinter` and removes the need for pkgconfig entirely.

jjhelmus: This does not remove the need for pkgconfig, it is still used to find other libraries such as libffi. Directly configuring Tcl/Tk on macOS could be done or pkgconfig can be used with this patch. There are tradeoffs with both options but I'm going to go with patching the pkgconfig file here so that no change is needed when upgrading Tcl/Tk versions.

jjhelmus: This is now done via an inline patch.

cpython-unix/build-tk.sh
jjhelmus: This is now done via an inline patch.

zanieb (approved):

jjhelmus:
Reported the missing Libs.private in tk.pc.in upstream, https://core.tcl-lang.org/tk/tktview/3e5d09f0e49755b2c6d9a7a5b94c990968bb9734

The modification to tcl.pc.in is specific to the use case of using the macOS SDK which does not provide pkg-config files. As such this does not warrant a report upstream.

PR #1181, opened 2026-07-18
CPython 3.15.0b4

jjhelmus:
* CPython 3.15.0b3 -> 3.15.0b4


jjhelmus:
Updated to newer version of `cmake` and `cc` which support Visual Studio 2026. The `windows-latest` runners now use VS 2026, which the previous locked versions cannot detect.

Issue #1186, opened 2026-07-22
valgrind diagnostic in uv python 3.15.0b4, but not locally built 3.15.0b4


jepler:
### Summary

While trying to diagnose an unrelated problem, I noticed that there's some kind of incorrect stack memory usage diagnosed by valgrind in this code:
```
$ uv run --python 3.15.0b4 valgrind --quiet --error-exitcode=13 --exit-on-first-error=yes python -c 'import _sha2; _sha2.sha512().digest()'
==999== Invalid write of size 8
==999==    at 0x1B9291A: sha512_update (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1B924A4: _Py_LibHacl_Hacl_Hash_SHA2_digest_512 (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1B9220E: SHA512Type_digest (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1A2796F: _TAIL_CALL_CALL (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1A51D49: _PyEval_Vector (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1A51A41: PyEval_EvalCode (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1AD4202: run_mod.llvm.10306309658545239246 (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1AD401C: _PyRun_String.llvm.10306309658545239246 (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1B7B54C: _PyRun_SimpleString (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
==999==    by 0x1B7B44C: Py_RunMain (in /root/.local/share/uv/python/cpython-3.15.0b4-linux-x86_64-gnu/bin/python3.15)
[...]

PR #1187, opened 2026-07-22
Skip BOLT optimization for HACL hash compressors

zanieb:
Optimized x86-64 GNU/Linux builds can perform invalid stack writes in the HACL SHA-2 and BLAKE2 compressors because BOLT's allocation combiner removes their required stack allocation while retaining frame-pointer-relative accesses below the red zone. Extend `-skip-funcs` for the affected compressor symbols so BOLT leaves these functions intact while continuing to optimize the rest of the interpreter.

See https://github.com/astral-sh/python-build-standalone/issues/1186

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -495,7 +495,11 @@ if [ -n "${CPYTHON_OPTIMIZED}" ]; then
         # https://github.com/python/cpython/issues/128514
         # https://github.com/python/cpython/pull/134642
         # https://github.com/llvm/llvm-project/issues/174191
-        BOLT_COMMON_FLAGS="-update-debug-sections -skip-funcs=RC4_options/1"
+        # BOLT's allocation combiner can remove required stack allocations in
+        # the HACL hash compressors, leaving frame-pointer-relative accesses
+        # below the x86-64 red zone. Match BOLT's `/N` suffix for local symbols.
+        # https://github.com/astral-sh/uv/issues/20618
+        BOLT_COMMON_FLAGS="-update-debug-sections -skip-funcs=RC4_options/1,sha256_update/.*,sha512_update/.*,PyBlake2_blake2b_compress/.*,PyBlake2_blake2s_compress/.*,update_block.*"
         BOLT_APPLY_FLAGS="${BOLT_COMMON_FLAGS} \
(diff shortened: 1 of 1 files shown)


cpython-unix/build-cpython.sh
```diff
@@ -495,7 +495,11 @@ if [ -n "${CPYTHON_OPTIMIZED}" ]; then
         # https://github.com/python/cpython/issues/128514
         # https://github.com/python/cpython/pull/134642
         # https://github.com/llvm/llvm-project/issues/174191
-        BOLT_COMMON_FLAGS="-update-debug-sections -skip-funcs=RC4_options/1"
+        # BOLT's allocation combiner can remove required stack allocations in
+        # the HACL hash compressors, leaving frame-pointer-relative accesses
+        # below the x86-64 red zone.
+        # https://github.com/astral-sh/uv/issues/20618
```
zanieb, line 501:
```suggestion
        # https://github.com/astral-sh/python-build-standalone/issues/1186
```

jjhelmus:
Switching to `-frame-opt=none` is another option. It is safer but could have an effect on performance.

zanieb:
I dunno why we'd turn that off broadly instead of skipping the relevant functions, I imagine it would affect performance?

jjhelmus:
> I dunno why we'd turn that off broadly instead of skipping the relevant functions, I imagine it would affect performance?

The identified functions are those in which we know that the stack redline is being violated. There could be others. Disabling frame optimization would prevent cases in other, yet unknown, functions.

jjhelmus:
> I imagine it would affect performance?

`-frame-opt=none` has a sizable effect on performance, from a quick fastmark benchmark, the score drops by ~1%.

jjhelmus:
Codex did not find any other functions with unsafe stack-frame removal resulting from BOLT outside the HACL related one. These function are implemented in ~assembly~ verified C so excluding them from additional optimization is reasonable.

jjhelmus (approved):

jjhelmus:
FWIW this issue only occurs because frame pointer/leaf from pointer are included (`-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer`). Prior to this change the functions use stack pointer based access which is detected by BOLT's allocation combiner.

PR #1190, opened 2026-07-23
Run stdlib tests via on-demand action

jjhelmus:
This is a rebase and modified version of #1068. 

Specifically this moves the running of the stdlib tests to a new action which is manually triggered. This new action downloads existing release artifacts and run the stdlib tests, optionally filtering by platform, Python version, or build options. 

From the original work:

This commit overhauls our ability to run the stdlib test harness and enables the stdlib test harness in CI for builds that we can run natively in CI.

Previously, `testdist.py` called a `run_tests.py` script that was bundled in the distribution. This script was simply a wrapper to calling `python -m test --slow-ci`. And `--slow-ci` currently expands to `--multiprocess 0 --randomize --fail-env-changes --rerun --print-slow --verbose3 -u all --timeout 1200`.

This commit effectively inlines `run_tests.py` into `testdir.py` as well as greatly expands functionality for running the test harness.

When enabling the stdlib test harness in CI as part of this commit, several test failures were encountered, especially in non-standard builds like `static` and `debug`. Even the `freethreaded` builds seemed to encounter a significant amount of failures (many of them intermittent), implying that the official CPython CI is failing to catch a lot of legitimate test failures.

We want PBS to run stdlib tests to help us catch changes in behavior. And we can only do that if the CI pass/fail signal is high quality: we don't want CI "passing" if there are changes to test pass/fail behavior.

Achieving this requires annotating all tests that can potentially fail. And then the test harness needs to validate that these annotations are accurate (read: that tests actually fail).

So this commit introduces a `stdlib-test-annotations.yml` file in the root directory. It contains rules that filter a build configuration and 3 sections that describe specific annotations:

1. Skip running the test harness completely. This is necessary on some builds that are just so broken it wasn't worth annotating tests because so many tests failed.
2. Exclude all tests within a given Python module. This is reserved for scenarios where importing the test module fails and causes most/all tests to fail. Again, a mechanism to short-circuit having to annotate every failing test.
3. Expected test failures. The most common annotation. These annotations describe individual tests or glob pattern matches of tests that are "expected" to fail. Entries can be annotated as "intermittent" or "dont-verify" to allow the test to pass without failing our test harness.

Most of the new code is in support of reading and applying these annotations.

At build time, we read the `stdlib-test-annotations.yml` file and derive a new `stdlib-test-annotations.json` file with only the active annotations matching the build configuration. This file is included in the build distribution as `python/build/stdlib-test-annotations.json`. It has to be JSON so the Python test harness runner is able to read the file using just the stdlib.

`test-distributions.py` has gained some new functionality, including the ability to run the stdlib test harness with raw arguments and emit a JUnit XML file with test results.

One of the things the test harness does now is attempt to ensure that tests annotated as failing actually fail. However, this isn't enforced for tests marked as "intermittent" or "dont-verify." You need an asynchronous mechanism looking at historical execution results to assess whether an "intermittent" test is such. We facilitate this by uploading a JUnit XML artifact with details of test execution. But the mining of historical test results is not implemented. (And I'm not sure if it is worth implementing.)

It took dozens of iterations to get a reliably working set of test annotations. There's just lots of variability across build configurations and Python versions. Despite best efforts, there's likely a few lingering intermittent failures that aren't yet annotated.

### DIFF
--- a/pythonbuild/testdist.py
+++ b/pythonbuild/testdist.py
@@ -3,14 +3,53 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 import argparse
+import concurrent.futures
+import dataclasses
+import enum
 import json
 import os
+import re
+import shlex
 import subprocess
 import tempfile
 from pathlib import Path
 from typing import Optional
 
+from junitparser import JUnitXml
+
+from .cpython import (
+    TEST_ANNOTATION_HARNESS_SKIP,
+    TEST_ANNOTATION_MODULE_EXCLUDE,
+    TEST_ANNOTATION_TEST_FAILURE,
+    TestAnnotation,
+    meets_python_minimum_version,
+)
 from .utils import extract_python_archive
+
+# Tests in the standard library that are slow.
+#
+# Roughly in order from slowest to fastest.
+STDLIB_SLOW_TESTS = [
+    "test_subprocess",
+    "test_signal",
+    "test_socket",
+    "test.test_multiprocessing_spawn.test_processes",
+    "test.test_multiprocessing_forkserver.test_processes",
+    "test_regrtest",
+    "test.test_concurrent_futures.test_process_pool",
+    "test_tarfile",
+    "test_socket",
+    "test_remote_pdb",
+    "test_threading",
+    "test.test_concurrent_futures.test_as_completed",
+    "test.test_multiprocessing_spawn.test_misc",
+    "test.test_asyncio.test_tasks",
+    "test_zipfile",
+    "test_pickle",
+]
+
+# Maximum wall run time of a single test before timing it out.
+TIMEOUT_SECONDS = 300
 
 
 def run_dist_python(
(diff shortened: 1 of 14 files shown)


pythonbuild/testdist.py
```diff
@@ -63,44 +108,472 @@ def run_custom_unittests(pbs_source_dir: Path, dist_root: Path, python_info) ->
...
+        parser.add_argument("--junit-xml", type=Path)
+        harness_options, _ = parser.parse_known_args(raw_harness_args)
+        if harness_options.junit_xml is not None:
+            junit_path = dist_root / harness_options.junit_xml
+        else:
+            # Insert before forwarded arguments, which may contain `--`.
+            args.extend(["--junit-xml", str(junit_path)])
+        args.extend(raw_harness_args)
+    else:
+        args.extend(
```
ewdurbin, line 200: you mentioned --slow-ci expanding to include `-u all`, is it intentional that we're not including _any_ resources here when we construct the arguments?

ewdurbin (approved):
overall looks good, i am just unclear on precisely what the intended behavior is for what resources are enabled when running the stdlib tests.

pythonbuild/testdist.py
jjhelmus: This seems unintentional and likely incorrect. Excluding resources will results in many more tests being skipped and not validated. 

I'll add this in, if the CI runners do not support certain resources those can be excluded explicitly.

PR #1191, opened 2026-07-23
Enable BOLT on aarch64-unknown-linux-gnu

jjhelmus:
Enable BOLT optimization during CPython build on the aarch64-unknown-linux-gnu platform.

To avoid https://github.com/llvm/llvm-project/issues/146541 the `_ctypes` extension is built as a shared module, causing the libffi trampoline code to exists in the extension rather than the `python` binary.

An alternative is to disable these static trampolines using `--disable-exec-static-tramp` when building libffi but this requires writable and executable memory segments which comes with security risks.

### DIFF
--- a/cpython-unix/build-cpython.sh
+++ b/cpython-unix/build-cpython.sh
@@ -504,7 +504,6 @@ if [ -n "${CPYTHON_OPTIMIZED}" ]; then
 -reorder-blocks=ext-tsp \
 -reorder-functions=cdsort \
 -split-functions \
--split-strategy=cdsplit \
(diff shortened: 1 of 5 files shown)


AlexPoone:
I tested the change on 3.14 and 3.15. arm-v8.2 CPUs. 
Both bolt arm64 CPython builds experience noticeable performance regressions, a bit but not large.

jjhelmus:
Just to confirm. Your tests show that the BOLT builds were slower/less performant than the baseline non-BOLT builds?

PR #1192, opened 2026-07-24
Add information about pbs performance

jjhelmus:
Add a document (`BENCHMARKS.md`) which discusses the performance on python-build-standalone in pyperformance benchmarks compared to other CPython distributions.

### DIFF
--- a/BENCHMARKS.md
+++ b/BENCHMARKS.md
@@ -1,0 +1,47 @@
+## python-build-standalone performance
+
+<img 
+    width="1820" height="1730" 
+    alt="Shows a chart with violin plots with benchmark results" 
+    src="https://github.com/user-attachments/assets/5499f9ce-ee02-4485-baa2-9d982fe457e6"
+/>
+
+python-build-standalone provides high-performance CPython distributions designed for compatibility across a broad range of Linux distributions, macOS releases, and Windows versions. It's builds incorporate compiler optimizations such as profile-guided optimization (PGO), link-time optimization (LTO), and, where appropriate, BOLT post-link binary optimization. Together, these techniques improve runtime performance while preserving portability.
+Benchmarks using the pyperformance suite show that python-build-standalone performs competitively with, and frequently outperforms, other widely used CPython distributions.
+
+
+### Figure details
+
+The figure above compares CPython 3.14.6 performance across several distributions using pyperformance. Each violin represents the distribution of per-benchmark mean runtime ratios between an alternative CPython distribution and python-build-standalone on the same platform and architecture. Ratios greater than 1 indicate that python-build-standalone was faster; ratios less than 1 indicate that the alternative distribution was faster.
+
+The horizontal axis uses a logarithmic scale. The vertical marker within each violin indicates the geometric mean of the runtime ratios, which is also expressed as a percentage beside each distribution.
+
+From top to bottom, the distributions shown are:
+* The Docker `python:3.14` image for x86-64, providing CPython 3.14.6.
+* A conda-forge Python 3.14.6 environment for `linux-64`.
+* The system Python 3.14.6 in a `fedora:44` x86-64 Docker container.
+* The system Python 3.14.6 in a debian:forky x86-64 Docker container.
+* CPython 3.14.6 from the Python.org macOS installer on an arm64 Mac.
+* A conda-forge Python 3.14.6 environment for `osx-arm64`.
+* CPython 3.14.6 installed through Homebrew on an arm64 Mac.
+* CPython 3.14.6 from the Python.org Windows installer for x86-64.
+* A conda-forge Python 3.14.6 environment for `win-64`.
+
+The reference interpreter for each comparison was the corresponding platform- and architecture-matched python-build-standalone CPython 3.14.6 distribution from the [`20260623` release](https://github.com/astral-sh/python-build-standalone/releases#release-20260623), installed using `uv`.
+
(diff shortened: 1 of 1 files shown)


jjhelmus:
Figure for BENCHMARKS.md

pbs-314-comparison

BENCHMARKS.md
```diff
@@ -0,0 +1,47 @@
+## python-build-standalone performance
+
+<img 
+    width="1820" height="1730" 
+    alt="Shows a chart with violin plots with benchmark results" 
+    src="https://github.com/user-attachments/assets/5499f9ce-ee02-4485-baa2-9d982fe457e6"
+/>
+
+python-build-standalone provides high-performance CPython distributions designed for compatibility across a broad range of Linux distributions, macOS releases, and Windows versions. It's builds incorporate compiler optimizations such as profile-guided optimization (PGO), link-time optimization (LTO), and, where appropriate, BOLT post-link binary optimization. Together, these techniques improve runtime performance while preserving portability.
```
zsol, line 9: nit: I feel like this might be appropriate for a blog post, but when you're already in `astral-sh/python-build-standalone/blob/main/BENCHMARKS.md` this doesn't add too much value. At the very least it's out of place, sitting below the chart. I'd start immediately with the "The figure above" paragraph

BENCHMARKS.md
```diff
@@ -0,0 +1,47 @@
...
+Benchmarks using the pyperformance suite show that python-build-standalone performs competitively with, and frequently outperforms, other widely used CPython distributions.
+
+
+### Figure details
+
+The figure above compares CPython 3.14.6 performance across several distributions using pyperformance. Each violin represents the distribution of per-benchmark mean runtime ratios between an alternative CPython distribution and python-build-standalone on the same platform and architecture. Ratios greater than 1 indicate that python-build-standalone was faster; ratios less than 1 indicate that the alternative distribution was faster.
```
zsol, line 15: maybe smuggle a link to https://en.wikipedia.org/wiki/Violin_plot ? I wasn't familiar with this until I looked it up

BENCHMARKS.md
```diff
@@ -0,0 +1,47 @@
...
+From top to bottom, the distributions shown are:
+* The Docker `python:3.14` image for x86-64, providing CPython 3.14.6.
+* A conda-forge Python 3.14.6 environment for `linux-64`.
+* The system Python 3.14.6 in a `fedora:44` x86-64 Docker container.
+* The system Python 3.14.6 in a debian:forky x86-64 Docker container.
+* CPython 3.14.6 from the Python.org macOS installer on an arm64 Mac.
+* A conda-forge Python 3.14.6 environment for `osx-arm64`.
+* CPython 3.14.6 installed through Homebrew on an arm64 Mac.
+* CPython 3.14.6 from the Python.org Windows installer for x86-64.
+* A conda-forge Python 3.14.6 environment for `win-64`.
```
zsol, line 28: should we include more precise versions that will allow people to reproduce the benchmark? e.g. the docker image hash, or the link to the python.org macos installer

BENCHMARKS.md
```diff
@@ -0,0 +1,47 @@
+## python-build-standalone performance
+
+<img 
+    width="1820" height="1730" 
+    alt="Shows a chart with violin plots with benchmark results" 
+    src="https://github.com/user-attachments/assets/5499f9ce-ee02-4485-baa2-9d982fe457e6"
+/>
```
zsol, line 7: should we include the raw data too, not just the image?

jjhelmus: A link to the script and raw data are now included.

BENCHMARKS.md
jjhelmus: Added a link to the macOS installer and a sentence at the end of this section directing to jjhelmus/cpython-benchmarks which holds the Dockerfiles and other supporting material.

The Dockerfiles were now referenced by hash/snapshot. This tried to reproduce what an end user would experience rather than a reproducible benchmarks.

BENCHMARKS.md
jjhelmus: Done

BENCHMARKS.md
jjhelmus: Agreed, removed this paragraph.

BENCHMARKS.md
```diff
@@ -0,0 +1,43 @@
...
+
+The figure above compares CPython 3.14.6 performance across several distributions using pyperformance. Each [violin](https://en.wikipedia.org/wiki/Violin_plot) represents the distribution of per-benchmark mean runtime ratios between an alternative CPython distribution and python-build-standalone on the same platform and architecture. Ratios greater than 1 indicate that python-build-standalone was faster; ratios less than 1 indicate that the alternative distribution was faster.
+
+The horizontal axis uses a logarithmic scale. The vertical marker within each violin indicates the geometric mean of the runtime ratios, which is also expressed as a percentage beside each distribution.
+
+From top to bottom, the distributions shown are:
+* The Docker `python:3.14` image for x86-64, providing CPython 3.14.6.
+* A conda-forge Python 3.14.6 environment for `linux-64`.
+* The system Python 3.14.6 in a `fedora:44` x86-64 Docker container.
+* The system Python 3.14.6 in a debian:forky x86-64 Docker container.
```
ntBre: Should this be code as well?


```suggestion
* The system Python 3.14.6 in a `debian:forky` x86-64 Docker container.
```

zsol (approved):

PR #1201, opened 2026-08-04
Tcl/Tk 9.0.4 on Windows

jjhelmus:
Update to the CPython provided Tcl/Tk 9.0.4 for 3.15+ on Windows.


jjhelmus:
This version is (likely) included in the Python.org installer and therefore should be included in the 3.15.0rc1 release.

PR #1205, opened 2026-08-04
Use a sparse checkout of the macOS SDKs

jjhelmus:
Use a sparse checkout of only the files needed during validations to speed up the checkout of the macOS SDKs repository.

A full check out typically takes between 2-3 minutes although occasionally the checkout can take much longer.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -171,6 +171,12 @@ jobs:
           repository: phracker/MacOSX-SDKs
           ref: master
           path: macosx-sdks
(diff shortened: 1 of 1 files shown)


jjhelmus:
Another method to speed up this step is to perform the checkout (sparse or full) once and cache the result for use in each job. Will look into this tomorrow.

jjhelmus:
A median time for the SDK checkout of 2m10s from the merge of #1202 to 11s for the first job here. 
If this takes <20 seconds it is likely not worth caching the results.

PR #1213, opened 2026-08-07
Enable static builds for the aarch64-linux-musl target

jjhelmus:
Enable static builds for the `aarch64-unknown-linux-musl` target.

### DIFF
--- a/cpython-unix/build-musl.sh
+++ b/cpython-unix/build-musl.sh
@@ -105,4 +105,17 @@ CFLAGS="${CFLAGS}" CPPFLAGS="${CPPFLAGS}" ./configure \
 make -j "$(nproc)"
 make -j "$(nproc)" install DESTDIR=/build/out
 
+if [ -n "${STATIC}" ] && [[ "$(clang -dumpmachine)" == aarch64-* ]]; then
+    # musl's linker wrapper appends libc after other arguments, including
+    # compiler-rt supplied through the target compiler flags. On aarch64,
+    # static libc uses compiler runtime builtins such as __multf3. GNU ld
+    # searches archives left to right and does not revisit earlier archives,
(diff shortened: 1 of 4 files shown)


cpython-unix/build.py
```diff
@@ -98,7 +98,9 @@ def add_target_env(env, build_platform, target_triple, build_env, build_options)
     extra_host_cflags = []
     extra_host_ldflags = []
 
-    # Add compiler-rt for aarch64-musl to resolve missing builtins
+    # aarch64 libgcc is built against glibc and initializes LSE atomics using
+    # __getauxval, which musl does not export.
+    # Use compiler-rt instead, which uses the public getauxval interface.
     if target_triple == "aarch64-unknown-linux-musl":
         extra_target_cflags.append("--rtlib=compiler-rt")
         extra_target_ldflags.append("--rtlib=compiler-rt")
```
jjhelmus, line 106: Note that this is where compiler-rt is added as a linker argument. This applies to all aarch64-musl builds, static and dynamic.

cpython-unix/build-musl.sh
```diff
@@ -105,4 +105,17 @@ CFLAGS="${CFLAGS}" CPPFLAGS="${CPPFLAGS}" ./configure \
 make -j "$(nproc)"
 make -j "$(nproc)" install DESTDIR=/build/out
 
+if [ -n "${STATIC}" ] && [[ "$(clang -dumpmachine)" == aarch64-* ]]; then
```
EliteTK:
```suggestion
if [[ $STATIC && $(clang -dumpmachine) == aarch64-* ]]; then
```

nitpick: This reads nicer and `[[` is generally easier to reason about than `[`.

EliteTK (approved):

PR #1216, opened 2026-08-11
Use raw syscalls for copy_file_range() and memfd_create()

daandemeyer:
glibc only grew the copy_file_range() and memfd_create() wrappers in 2.27, and CPython compiles os.copy_file_range() and os.memfd_create() out when the libc it is built against lacks them. We worked around that by forcing the configure checks on and weak linking the wrappers, which kept the functions out of the os module whenever the runtime glibc was older than 2.27, even on kernels implementing the syscalls.

Backport https://github.com/python/cpython/pull/155520 instead, which calls the wrappers when they exist and issues the raw syscalls when they don't, so both functions work on any sufficiently new kernel regardless of the glibc in use. The UAPI header overlay already provides the __NR_ constants and the MFD_ flags for all glibc targets. A single patch covers 3.10 through 3.15, so the weak linking patches and the 3.10 specific configure patch go away.

The distribution tests now assert that both functions are always present on Linux GNU targets instead of tying their availability to the runtime libc.


jjhelmus:
This seems like a reasonable approach and once merged upstream I'm in favor of switching. 
I'm going to expand the CI matrix to get more complete test coverage.

daandemeyer:
Should have been merged upstream by the time you read this message.

PR #1217, opened 2026-08-12
skip llvm-strip on signed DLLs

ethernet8023:
when packaging a `uv python` installed python into a .msix app for the microsoft app store, i hit some `signtool` failures with error `0x800700C1`. 

these are caused by `AppxSIP` going thru every binary in your package and validating them in some way.

`llvm-strip` removes the signature from a bunch of DLLs that come with these python buidls, but it keeps the certificate table entry in the PE optional header , which makes the entry point past the end of the file.

this is *exactly* the bug from https://github.com/astral-sh/python-build-standalone/issues/855 that https://github.com/astral-sh/python-build-standalone/pull/856 solved. i'm extending it to more DLLs and adding a comment explaining it.

a more permanent fix might be detecting signatures on files and skipping if we see them? let me know if you'd rather that, i could wireup detecting signatures and/or zipfs to get rid of the hardcoded list that'll probably regress as versions change what DLLs are shipped.

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -636,11 +636,27 @@ pub fn convert_to_stripped<W: Write>(
                 | FileKind::Pe32
                 | FileKind::Pe64)
         ) {
-            // Skip stripping MSVC runtime DLLs and Tcl/Tk DLLs containing ZIPFS data.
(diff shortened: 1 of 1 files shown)


jjhelmus:
This makes sense. These DLLs are coming pre-build (and signed) from CPython via https://github.com/python/cpython-bin-deps.

PR #1220, opened 2026-08-18
Add ppc64le power9, power10, and power11 optimized build targets

### LINKED ISSUE #1215: Additional arch binaries for ppc64le
Creating this issue for discussion as we're working on some PRs. 

The existing standalone python binary is built optimized for the Power 9 family of processors, since it uses the gcc defaults inherited in the cross-compile base image (gcc6). We're looking at how to provide options for Power10 , Power11 and future users. Just switching the `-mtune` parameter and rebuilding isn't ideal since 1) it wouldn't be backwards compatible, and 2)  power10 & power11 values were added in GCC 10+ and GCC 13+ respectively. 

I'm thinking the simplest thing to do create processor binaries as is done for x86, e.g.
- ppc64le-unknown-linux-gnu
- ppc64le_power9-unknown-linux-gnu
- ppc64le_power10-unknown-linux-gnu
- ppc64le_power11-unknown-linux-gnu

Next we would have to also submit a PR to https://github.com/astral-sh/uv so that it checks the host version at install time and chooses the right processor-specific version. 

Does this sound reasonable? Some of the team are working on changes to that affect but thought we should float it here first. 

Also this may affect how bazel and other toolchains pull in the power python binary. It's been a while since I personally worked in that space so may no longer be a concern.

veenious:
Adds three new ppc64le architecture variant targets, mirroring the existing
x86_64_v2/v3/v4 pattern:

- `ppc64le_power9-unknown-linux-gnu` — compiled with `-mcpu=power9 -mtune=power9`
- `ppc64le_power10-unknown-linux-gnu` — compiled with `-mcpu=power10 -mtune=power10`
- `ppc64le_power11-unknown-linux-gnu` — compiled with `-mcpu=power11 -mtune=power11`

The existing generic `ppc64le-unknown-linux-gnu` artifact is untouched.

Power9 support is based on Christy Norman's work (clnperez/python-build-standalone#1).
Power10 uses Debian Bookworm's GCC 12. Power11 uses GCC 14 from Debian Trixie because
GCC 12 does not support `-mcpu=power11`.

> **Note:** `ppc64le_power10` and `ppc64le_power11` require glibc ≥ 2.36 at runtime
> (enforced in `validation.rs`). This is a consequence of using Debian Bookworm as the
> build base. The existing `ppc64le-unknown-linux-gnu` and `ppc64le_power9` targets
> retain glibc 2.17 compatibility.

Closes #1215

### DIFF
--- a/cpython-unix/build.cross-ppc64le_power10.Dockerfile
+++ b/cpython-unix/build.cross-ppc64le_power10.Dockerfile
@@ -1,0 +1,54 @@
+# Debian Bookworm.
+FROM docker.io/library/debian@sha256:6bc30d909583f38600edd6609e29eb3fb284ab8affce8d0389f332fc91c2dd91
+LABEL org.opencontainers.image.authors="Gregory Szorc <gregory.szorc@gmail.com>"
+
+RUN groupadd -g 1000 build && \
+    useradd -u 1000 -g 1000 -d /build -s /bin/bash -m build && \
+    mkdir /tools && \
+    chown -R build:build /build /tools
+
+ENV HOME=/build \
+    SHELL=/bin/bash \
+    USER=build \
+    LOGNAME=build \
+    HOSTNAME=builder \
+    DEBIAN_FRONTEND=noninteractive
+
+CMD ["/bin/bash", "--login"]
+WORKDIR '/build'
+
+RUN printf '%s\n' \
+      'deb https://deb.debian.org/debian bookworm main' \
+      'deb https://deb.debian.org/debian bookworm-updates main' \
+      'deb https://deb.debian.org/debian-security bookworm-security main' \
+    > /etc/apt/sources.list && \
+    ( echo 'APT::Get::Assume-Yes "true";'; \
+      echo 'APT::Install-Recommends "false";'; \
+      echo 'Acquire::Retries "5";'; \
+      echo 'Acquire::https::Verify-Peer "false";'; \
+    ) > /etc/apt/apt.conf.d/99cpython-portable && \
+    rm -f /etc/apt/sources.list.d/* && \
+    apt-get update
+
+# Host building.
+RUN apt-get install \
+    bzip2 \
+    gcc \
+    g++ \
+    libc6-dev \
+    libffi-dev \
+    make \
+    patch \
+    perl \
+    pkg-config \
+    tar \
+    xz-utils \
+    unzip \
+    zip \
+    zlib1g-dev
+
+# Cross-building.
+RUN apt-get install \
+    g++-powerpc64le-linux-gnu \
+    gcc-powerpc64le-linux-gnu \
+    libc6-dev-ppc64el-cross
--- a/cpython-unix/build.cross.Dockerfile
+++ b/cpython-unix/build.cross.Dockerfile
@@ -1,5 +1,5 @@
 # Debian Stretch.
-FROM debian@sha256:cebe6e1c30384958d471467e231f740e8f0fd92cbfd2a435a186e9bada3aee1c
+FROM docker.io/library/debian@sha256:cebe6e1c30384958d471467e231f740e8f0fd92cbfd2a435a186e9bada3aee1c
 LABEL org.opencontainers.image.authors="Gregory Szorc <gregory.szorc@gmail.com>"
 
 RUN groupadd -g 1000 build && \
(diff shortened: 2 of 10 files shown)


src/validation.rs
```diff
@@ -201,6 +202,10 @@ static GLIBC_MAX_VERSION_BY_TRIPLE: Lazy<HashMap<&'static str, version_compare::
             "ppc64le-unknown-linux-gnu",
             version_compare::Version::from("2.17").unwrap(),
         );
+        versions.insert(
+            "ppc64le_power9-unknown-linux-gnu",
+            version_compare::Version::from("2.17").unwrap(),
+        );
```
jjhelmus, line 211: There are no validators for power10 and 11. A minimum glibc version should be defined and the allowed set of libraries.

jjhelmus (changes requested):

veenious:
Validation update (UBI 10):

I built CPython 3.13.15 `noopt` artifacts for the POWER9, POWER10, and POWER11 targets and tested each artifact on matching ppc64le hardware using `registry.access.redhat.com/ubi10/ubi` (glibc 2.39).

All three passed:
- `python --version` reported 3.13.15
- imports: `ssl`, `sqlite3`, `zlib`, `json`, `math`, `hashlib`, and `ctypes`
- bundled `pip` (26.2.1)
- a decimal-pi calculation to 10,000 digits

Measured pi benchmark times:
- POWER9: 2.334s
- POWER10: 2.021s
- POWER11: 1.905s

The artifacts require glibc >= 2.36 and therefore were tested in UBI 10 rather than directly on the RHEL 9.8 hosts (glibc 2.34). Per discussion, I am focusing support and validation on the UBI 10 experience.

src/validation.rs
veenious: Added Power10 and Power11 entries to all validation maps, including recognized triples, a GLIBC 2.36 maximum, allowed ELF libraries, platform tags, and ELF machine validation.

veenious:
## Hardware test results — Python 3.13.15 on real P9/P10/P11 hardware

All three binaries were cross-compiled from x86 (commit cbcc593) and tested inside
`registry.access.redhat.com/ubi10/ubi` (glibc 2.39) containers on RHEL 9.8 hosts.

### Environment

| Machine | CPU | Kernel | Host glibc |
|---|---|---|---|
| P9 | POWER9 (architected), altivec supported | 5.14.0-687.38.1.el9_8.ppc64le | 2.34 |
| P10 | POWER10 (architected), altivec supported | 5.14.0-687.38.1.el9_8.ppc64le | 2.34 |
| P11 | Power11 (architected), altivec supported | 5.14.0-687.12.1.el9_8.ppc64le | 2.34 |

### Smoke tests (all passed)

All three binaries passed: Python version check, `ssl`, `sqlite3`, `zlib`, `json`,
`math`, `hashlib`, `ctypes` imports, and `pip` invocation.

### Performance — Pi to 10,000 decimal digits

| Binary | Time | vs Power9 |
|---|---|---|
| `ppc64le_power9-unknown-linux-gnu` | 37.4 s | baseline |
| `ppc64le_power10-unknown-linux-gnu` | 31.4 s | **+16% faster** |
| `ppc64le_power11-unknown-linux-gnu` | 29.7 s | **+21% faster** |

### glibc requirement

Power9 requires glibc ≥ 2.17 (same as the generic baseline). Power10 and Power11
require glibc ≥ 2.36 due to the Bookworm build base. This is enforced in
`validation.rs`.

cpython-unix/build.cross.Dockerfile
```diff
@@ -17,12 +17,11 @@ ENV HOME=/build \
 CMD ["/bin/bash", "--login"]
 WORKDIR '/build'
 
-# Stretch stopped publishing snapshots in April 2023. Last snapshot
-# is 20230423T032533Z. But there are package authentication issues
-# with this snapshot.
-RUN for s in debian_stretch debian_stretch-updates debian-security_stretch/updates; do \
-      echo "deb http://snapshot.debian.org/archive/${s%_*}/20221105T150728Z/ ${s#*_} main"; \
-    done > /etc/apt/sources.list && \
```
jjhelmus, line 25: The snapshot release should not be removed. We want reproducible images when possible

cpython-unix/build.cross-ppc64le_power10.Dockerfile
```diff
@@ -0,0 +1,54 @@
+# Debian Bookworm.
+FROM docker.io/library/debian@sha256:6bc30d909583f38600edd6609e29eb3fb284ab8affce8d0389f332fc91c2dd91
+LABEL org.opencontainers.image.authors="Gregory Szorc <gregory.szorc@gmail.com>"
```
jjhelmus, line 3: This would need to have a different author

PR #1232, opened 2026-08-25
Validate GNU build IDs in Linux artifacts

jjhelmus:
Require executables and shared libraries to contain a nonempty GNU build ID to support debuginfod.

Follow up to #1230


EliteTK (approved):
This seems fine.

We could use [`object::read::Object::build_id`](https://docs.rs/object/latest/object/read/trait.Object.html#method.build_id) here, but the current approach should work for our toolchains.

We do strip after this, and we don't check after that. Stripping shouldn't remove this but who knows...

PR #1235, opened 2026-08-26
Refactor derive_setup_local function in the pythonbuild.cpython module

jjhelmus:
Refactor `derive_setup_local` into smaller, typed helpers while preserving existing extension module configuration and linkage behavior.

* Separate parsing, classification, validation, and setup generation.
* Dataclass to group extension classification and CPython module information.
* Parse module configuration files from the CPython source archives rather than from lines.
* Simplify static/shared linkage selection and Setup.local generation.

### DIFF
--- a/pythonbuild/cpython.py
+++ b/pythonbuild/cpython.py
@@ -5,6 +5,9 @@
 import pathlib
 import re
 import tarfile
+from collections.abc import Iterable
+from dataclasses import dataclass
+from typing import IO
 
 import jsonschema
 import yaml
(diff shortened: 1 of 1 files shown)


pythonbuild/cpython.py
```diff
@@ -447,63 +387,364 @@ def derive_setup_local(
...
+
+    if not parsed:
+        raise Exception("we should always parse a setup line we generated")
+
+    # makesetup interprets lines containing = as configuration options. Move
+    # -Dname=value defines into Makefile overrides for legacy Python builds.
+    if not use_setup_stdlib:
+        for match in define_pattern.finditer(parsed["line"]):
+            for obj_path in sorted(parsed["posix_obj_paths"]):
+                extra_cflags.setdefault(bytes(obj_path), []).append(match.group(0))
```
ewdurbin, line 681: a little nit, but passing in the `extra_cflags` just to mutate them in the function feels a little surprising given the name. should this function return the flags that need added and let the caller apply them?

ewdurbin (approved):

pythonbuild/cpython.py
jjhelmus: Good catch. I spotted this in one of the run throughs but never got a fix in. 
I'll address this in #1237 as that shuffles this block around further.

PR #1237, opened 2026-08-26
Derive PYTHON.json from build artifacts

jjhelmus:
* Generate Python 3.12+ extension metadata from configured Setup files, config.c, Makefile variables, and linker flags.
* Include object files from internal static archives such as HACL.
* Preserve legacy generation for Python 3.10/3.11.

Follow up to work in #1180

### DIFF
--- a/pythonbuild/cpython.py
+++ b/pythonbuild/cpython.py
@@ -565,8 +565,11 @@ def configured_extension_modules(
         for word in setup_line.split():
             if word.endswith(b".a") and not word.startswith(b"-"):
                 archive = word.decode("ascii")
-                obj_paths = archive_objects.get(archive, ())
-                archive_obj_paths.update(obj_paths)
+                if archive not in archive_objects:
+                    raise ValueError(
+                        f"extension {name} references unknown static archive {archive}"
+                    )
+                archive_obj_paths.update(archive_objects[archive])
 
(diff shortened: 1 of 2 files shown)


pythonbuild/cpython.py
```diff
@@ -392,6 +393,189 @@ def _parse_setup(
...
+        setup_line = setup_lines[name]
+        linker_flags = config_vars.get(f"MODULE_{name.upper()}_LDFLAGS", "")
+        if linker_flags:
+            setup_line += f" {linker_flags}".encode("ascii")
+
+        archive_obj_paths: set[pathlib.Path] = set()
+        for word in setup_line.split():
+            if word.endswith(b".a") and not word.startswith(b"-"):
+                archive = word.decode("ascii")
+                obj_paths = archive_objects.get(archive, ())
```
ewdurbin: in what scenarios do we expect to not have a value for a given `archive` value here? should we handle those explicitly or fail loudly?

ewdurbin (approved):

pythonbuild/cpython.py
jjhelmus: Yes, this should be a loud failure. These are internal CPython archives who objects need to be included. External archives are already excluded using the `word.startswith(b"-")` check.

PR #1241, opened 2026-08-27
Build aarch64 Linux distributions on Debian 13 with a Debian 9 sysroot

jjhelmus:
Use modern Debian 13 (Trixie) build containers to build Linux aarch64 distribution with a sysroot from Debian 9 (Stretch) to preserving compatibility with glibc 2.17.

* Update aarch64 containers to Debian Trixie
* Create an aarch64 sysroot from Debian Stretch cross-development packages.
* Configure GNU builds to compile and link against the legacy sysroot.
* Remove container-specific sysroot flags from sysconfig.


cpython-unix/build-cpython.sh
```diff
@@ -1098,6 +1098,12 @@ if xcode_path:
 # -fdebug-default-version is Clang only. Strip so compiling works on GCC.
 replace_in_all("-fdebug-default-version=4", "")
 
+# Target sysroots only exist in the build container. Keeping their paths in
+# sysconfig would make downstream extension builds search a nonexistent root.
+for flag in os.environ.get("EXTRA_TARGET_CFLAGS", "").split():
+    if flag.startswith("--sysroot="):
+        replace_in_all(flag, "")
```
EliteTK, line 1105: This actually misses the contents of `lib/python3.14/_sysconfig_vars_*.json`, it's just metadata though, `sysconfig` doesn't read it.

EliteTK (approved):

cpython-unix/build-cpython.sh
jjhelmus: None of the sysconfig replacements/edits are done to the .json file. Modifying this file similar to sysconfig can be done as a follow to also address the other edits but this is not needed for functionality.

PR #1251, opened 2026-09-02
Verify LLVM bootstrap archives before extraction

zsol:
The Rust LLVM bootstrap currently downloads and extracts archives without checking the pinned size or SHA-256.

Verify fresh and cached archives before extraction or reuse. Reject HTTP errors and integrity mismatches, and replace the cache only after successful extraction.

Stacked on #1250.

### DIFF
--- a/src/release.rs
+++ b/src/release.rs
@@ -2,9 +2,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use anyhow::Context;
+use anyhow::{Context, ensure};
 use futures::StreamExt;
 use serde::Deserialize;
+use sha2::{Digest, Sha256};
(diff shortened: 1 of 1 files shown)


src/release.rs
```diff
@@ -780,54 +784,113 @@ fn llvm_download(os: &str, arch: &str) -> Result<&'static Download> {
...
+            ensure!(
+                size <= download.size,
+                "LLVM archive exceeds expected size of {} bytes",
+                download.size
+            );
+            tarball_file.write_all(&chunk).await?;
+        }
+        tarball_file.flush().await?;
+        drop(tarball_file);
+        verify_llvm_archive(std::fs::File::open(&tarball_path)?, download)?
```
zanieb, line 878: I'm surprised we're not hashing as we write chunks?

zanieb (approved):
not the most performant way to do it but it seems fine here

PR #1253, opened 2026-09-04
use mmdebstrap to create aarch64 sysroot

jjhelmus:
Use mmdebstrap to create the aarch64 Stretch based sysroot. This avoids requiring a Debian 9 (Stretch) image entirely. Additionally this uses standard Debian aarch64 packages in the sysroot rather than `-cross` packages.


ewdurbin (approved):
I think this is once again approximately fine, but bypassing all the gpg faff for pinned metadata as proposed in https://github.com/astral-sh/python-build-standalone/pull/1254#discussion_r3961937066 seems less bad

PR #1254, opened 2026-09-04
Use Debian 13 build images for x86-64 Linux

jjhelmus:
Use modern Debian 13 (Trixie) build containers to build Linux x86-64 distribution with a sysroot from Debian 8 (Jessie) to preserving compatibility with glibc 2.17.

* Update x86-64 containers to Debian Trixie
* Create a x86-64 sysroot from Debian Jessie packages using mmdebstrap with HTTPS and signature validation.
* Configure GNU builds to compile and link against the legacy sysroot.

The `--sysroot` flags are removed from sysconfig based off addition done in #1241.

This similar to the approach in #1241 and #1253


cpython-unix/build.debian13-x86_64.Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    --architectures=amd64 \
+    --keyring=/usr/share/keyrings/debian-archive-removed-keys.gpg \
+    --aptopt='Acquire::Check-Valid-Until "false"' \
+    --aptopt='Acquire::Retries "5"' \
+    --aptopt='Apt::Key::gpgvcommand "/usr/libexec/mmdebstrap/gpgvnoexpkeysig"' \
+    --include=libc6,libc6-dev,linux-libc-dev,symlinks \
+    jessie /sysroot \
+    'deb [signed-by=126C0D24BD8A2942CC7DF8AC7638D0442B90D010!] https://snapshot.debian.org/archive/debian/20230322T152120Z/ jessie main' \
+    'deb [signed-by=126C0D24BD8A2942CC7DF8AC7638D0442B90D010!] https://snapshot.debian.org/archive/debian/20230322T152120Z/ jessie-updates main' \
+    'deb [signed-by=D21169141CECD440F2EB8DDA9D6D8F6BC857C906!] https://snapshot.debian.org/archive/debian-security/20230322T152120Z/ jessie/updates main'
```
ewdurbin, line 48: If you add in the similar checksum validated pull from the other PR like:

```
{% for repository, suite, sha256 in [
    ('debian', 'jessie', '7240a1c6ce11c3658d001261e77797818e610f7da6c2fb1f98a24fdbf4e8d84c'),
    ('debian', 'jessie-updates', 'f61f27bd17de546264aa58f40f3aafaac7021e0ef69c17f6b1b4cd7664a037ec'),
    ('debian-security', 'jessie/updates', '3da1205c671e38db711a76403b27f1b3e0b84766edcf717a4b9daea9d4c693b2'),
] %}
ADD --checksum=sha256:{{ sha256 }} \
    https://snapshot.debian.org/archive/{{ repository }}/20230322T152120Z/dists/{{ suite }}/main/binary-amd64/Packages.gz \
    /pinned-jessie-lists/snapshot.debian.org_archive_{{ repository }}_20230322T152120Z_dists_{{ suite | replace('/', '_') }}_main_binary-amd64_Packages.gz
{% endfor %}
```
you could modify the mmdebstrap call like so and drop gpgv from the apt install

```suggestion
RUN mmdebstrap --variant=extract --mode=root \
    --skip=chroot/mount \
    --skip=update \
    --architectures=amd64 \
    --aptopt='Acquire::Retries "5"' \
    --setup-hook='cp /pinned-jessie-lists/* "$1/var/lib/apt/lists/"' \
    --include=libc6,libc6-dev,linux-libc-dev,symlinks \
    jessie /sysroot \
    'deb [trusted=yes] https://snapshot.debian.org/archive/debian/20230322T152120Z/ jessie main' \
    'deb [trusted=yes] https://snapshot.debian.org/archive/debian/20230322T152120Z/ jessie-updates main' \
    'deb [trusted=yes] https://snapshot.debian.org/archive/debian-security/20230322T152120Z/ jessie/updates main'
```

ewdurbin (approved):
👍🏼 with an alternative that goes back to trusted pinning over ignoring expired gpg keys.

i guess you _could_ have both?

EliteTK (approved):
For aarch64 we start with an older image rather than mmdebootstrapping. It seemed cleaner, although I think it wouldn't be as simple here... Just feels weird to deviate.

cpython-unix/build.debian13-x86_64.Dockerfile
jjhelmus: I updated this block to check the SHA256 of the Release/InRelease files of the snapshot, which have scope one above the Packages.gz files, while retaining the validation of the expired signature. Seems like it is possible to have both protections. 

This also embeds the commands to manually obtains the SHA256 for these files if needed.

Issue #746, opened 2025-08-11
Fails to open `libtk8.6.so` on Ubuntu 22.04 when bundled with PyInstaller


johnthagen:
We have a PyInstaller binary that we are building that opens a `tk.TopLevel` dialog.

When the executable is launched on Ubuntu 22.04, at the line

```py
import tkinter as tk
```

The follow exception is raised:

```
  File "PyInstaller/loader/pyimod02_importers.py", line 457, in exec_module
  File "tkinter/__init__.py", line 38, in <module>
ImportError: libtk8.6.so: cannot open shared object file: No such file or directory
```

On Ubuntu 2**4**.04, it _does_ work.

CC @geofft , perhaps related to

- #676

### Motivation

We'd like to make our binaries as portable as possible, and `python-build-standalone` has a low GLIBC requirement that helps with this.

### Environment

- Python 3.13.6 (`python-build-standalone`)
-  pyinstaller v6.15.0
-  pyinstaller-hooks-contrib v2025.8
- Build environment: `rockylinux:8` container
  - `uv` 0.8.8

johnthagen:
If I switch from `python-build-standalone` to a `python:3.13-bullseye` system interpreter, then the executable works on both Ubuntu 22.04 and 24.04.

We'd prefer to the `python-build-standalone`, since it's the simplest way (in our experience) to get the latest Python releases that also have a very old minimum GLIBC requirement.

geofft:
I think this is going to be fixed by #745. The builds there should work and if you're bored I'd appreciate knowing if they do, but I likely will not have time to fix the tests and release this week. (See also https://github.com/pyinstaller/pyinstaller/issues/9204#issuecomment-3176436360 and the rest of that issue)

johnthagen:
@geofft I wanted to check back in on this. I saw [PyInstaller 6.16.0 was released](https://github.com/pyinstaller/pyinstaller/releases/tag/v6.16.0) and included:

- https://github.com/pyinstaller/pyinstaller/pull/9212
- https://github.com/pyinstaller/pyinstaller/pull/9218
- https://github.com/pyinstaller/pyinstaller/pull/9226

I was curious if I should expect these fixes to fix the TK issue listed here, or if further `python-build-standalone` fixes are also required?

johnthagen:
[...]

PR #1258, opened 2026-09-09
include a origin relative RPATH in _tkinter

jjhelmus:
`_tkinter` on Linux includes a ORIGIN relative RPATH allowing the Tcl/Tk libraries to be located without relying on the interpreter or libpython providing the search path. 

In typically workflows this is not necessary but certain tools scan native library and will fail to include the required libraries if they do not declare a search path, for example pyinstaller (#746).

c.f. #745, #742


geofft (approved):
Seems fine to unconditionally use an rpath here instead of doing the thing I was trying to do in #745 to do $ORIGIN-relative DT_NEEDEDs.

PR #1261, opened 2026-09-11
test out windows-11-vs2026-arm

jjhelmus:
testing out effect of switching to VS2026 in the aarch64 windows runner for the upcoming migration:

https://github.blog/changelog/2026-08-20-windows-11-arm64-vs2026-image-generally-available/


jjhelmus:
@vadikmironov Thanks for looking at the upgrade. I've been doing some of my own investigation. Rather than using the v145 toolset for all arm builds I think the v143 toolset should be selected using the VS 2026 vcvars script. Currently the platform toolset is coupled to the Visual Studion version but this is not necessary. I am working on a PR that decouples these two options.

jjhelmus:
#1266 implement decoupling of the Visual Studio version from the platform toolset / VC tools version. Some work in #1265 is needed here to internalize the environment setup which was previously done in the GHA.

jjhelmus:
#1272 has fully working builds for arm64 on the new image. Closing this PR.

Issue #26, opened 2019-11-27
Add debug support for MSVC builds


jayvdb:
I dont see `--debug` support in the Windows build scripts.  Debug builds are supported by the CPython PCbuild system. https://github.com/python/cpython/blob/master/PCbuild/readme.txt

indygreg:
Yes, this would be nice. The hacks to the Windows build system are ugly and the lack of debug support is mainly a side-effect of me wanting to preserve more of my sanity by not having to hack on the Visual Studio builds any more than necessary.

vadikmironov:
Windows debug builds are now published, starting with release 20260924 (https://github.com/astral-sh/python-build-standalone/releases/tag/20260924): debug for 3.10+ on i686 and x86_64 and 3.11+ on aarch64, plus freethreaded+debug for 3.13+ on all three (-full archives only).

@jjhelmus, thanks a lot for #1283 and for reviewing and merging #1262 / #1287! I think the issue can be closed now, could you close it when you get a chance?

jjhelmus:
Thanks for all your work on getting this created @vadikmironov  🎆.

PR #1262, opened 2026-09-12
Add debug support for MSVC builds - full change PR

vadikmironov:
This PR is a full implementation of debug support for Windows build (addresses #26 in full) and combines three commits together only for the purposes of review. Should maintainers have a preference to proceed with it, happy to assist, but otherwise I will be raising a set of three separate PRs for each commit to be discussed and merged.

**1. Add a debug build option for Windows** (`cpython-windows/build.py`, `docs/building.rst`)

- `debug` is added to the option set beside `noopt` and `pgo`, with the same ordering as Unix build already uses. Same as with Unix, this removes illegal combinations like `debug+pgo`. The set is renamed from `optimizations` to `options` also inline with the Unix build.
- `configuration="Debug"` is passed to msbuild and `artifact_config` accordingly. `PC/layout` gets `--debug` and that drives the `_d` artifact suffixes.
- executables, extension modules and dependency libraries pick up the `_d` suffix `$(PyDebugExt)` produces
- the tail-calling interpreter is not enabled for Debug (first finding below, comments embedded in the patch)
- `docs/building.rst` updated with the debug option and the debug CRT requirement

**2. Accept a Windows debug distribution in the validator and disttests** (`src/validation.rs`, `pythonbuild/disttests/__init__.py`)

- the PE allow list gains the debug CRT (`ucrtbased.dll`, `VCRUNTIME140D.dll`, `VCRUNTIME140_1D.dll`), the `_d` interpreter DLLs and `sqlite3_d.dll`
- the `abiflags` check consults `EXT_SUFFIX` on Windows (comments as to why are embedded inn the patch)
- `test_ssl_with_keylogfile` is skipped on Windows debug builds following CPython's suite skipping keylog there

**3. Build the Windows debug configurations in CI** (`ci-targets.yaml`)

- each Windows target has `pgo` and `freethreaded+pgo`; each gains `debug` and `freethreaded+debug`

## Evidence

Every configuration added in the third patch was built, validated and tested: 26 jobs, VS 2022 everywhere except x64 3.15 on VS 2026, arm64 on `windows-11-arm` through `vcvarsamd64_arm64.bat` (I see #1261 raised and I will try my testbed on it too as it may be relevant for my changes). `debug` for 3.10 to 3.15 on x64, x86 and (from 3.11) arm64; `freethreaded+debug` for 3.13 to 3.15 on all three. All green: https://github.com/vadikmironov/omniglot-bazel-starter/actions/runs/34095975591

Each job ran exactly what `windows.yml` runs: `validate-distribution` and then `test-distribution.py` through the distribution's own interpreter (22 tests, 12 skipped, OK on every job). The archives carry `python314_d.dll` and `libs/python314_d.lib`, `python_d.exe` next to the canonical `python.exe` copy the existing code makes, `_d.cp314-win_amd64.pyd` as `EXT_SUFFIX`, and `Py_DEBUG=1` in the config vars; the free-threaded variant adds the `t` as expected. An embedder compiled `/MDd` against each links with only a `/LIBPATH`, and `sys.gettotalrefcount()` works at runtime.

## Four gotchas that are commented in the patches

**The tail-calling interpreter cannot be built in Debug.** A guaranteed tail call needs optimisations which are removed by `/Od`. Microsoft documents `[[msvc::musttail]]` as requiring `/O2` (https://learn.microsoft.com/en-us/cpp/cpp/attributes#msvcmusttail), and CPython's gh-148047 shows the same flood of errors for `build.bat -c debug --tail-call-interp`. It is off for Debug.

**`Py_DEBUG` and `ABIFLAGS` reach Windows config vars only from 3.14.** 3.13's `_sysconfig.config_vars()` supplies `EXT_SUFFIX`, `SOABI` and `Py_GIL_DISABLED`; 3.14 adds `Py_DEBUG` and emulates `ABIFLAGS` while keeping lowercase `abiflags` empty on Windows (https://github.com/python/cpython/blob/v3.14.7/Lib/sysconfig/__init__.py#L407). Validator reads `EXT_SUFFIX` instead.

**From 3.14 a free-threaded Windows install no longer declares itself.** 3.14 ships a static `PC/pyconfig.h` that only normalises an externally supplied `Py_GIL_DISABLED`, so an embedder that does not pass `/DPy_GIL_DISABLED=1` gets a `#pragma comment(lib, ...)` naming the non-free-threaded import library. No changes per se and it applies to existing free-threaded builds already, but I will investigate the right course of actions separately.

**`SSLContext.keylog_filename` is unsupported on Windows debug builds.** CPython's suite skips every keylog test there (https://github.com/python/cpython/blob/v3.11.16/Lib/test/test_ssl.py#L4827), and from 3.12 the call raises `NotImplementedError` under `_DEBUG` (https://github.com/python/cpython/blob/v3.12.14/Modules/_ssl/debughelpers.c#L168, from python/cpython#131839). `test_ssl_with_keylogfile` is skipped on every Windows debug build same CPython.

## Limitations

- a debug distribution needs the debug CRT: `ucrtbased.dll` and `vcruntime140d.dll` are not redistributable by Microsoft. It only runs where Visual Studio is installed and it's exactly the same for CPython's debug binarie.
- 3.10 has no arm64 configuration, following what `ci-targets.yaml` lists for that version
- release-ABI wheels install into a debug interpreter but do not load, since Windows extension suffixes are `_d`-prefixed; CPython behaviour
- `crt_features` stays `vcruntime:140`, though the debug binaries link the non-redistributable runtime. This is another open point and happy to tweak it further.

## CI

Once the third commit is merged, it will add 26 jobs to every push to main, 15 to 26 minutes each on free runners. Again, happy to drop or amend it if that is too much.


cpython-windows/build.py
```diff
@@ -1211,12 +1218,13 @@ def find_additional_dependencies(project: pathlib.Path):
     else:
         raise Exception("unhandled architecture: %s" % arch)
 
+    debug_suffix = "_d" if config == "Debug" else ""
     if freethreaded:
-        abi_tag = ".cp%st-%s" % (python_majmin, abi_platform)
-        lib_suffix = "t"
+        abi_tag = "%s.cp%st-%s" % (debug_suffix, python_majmin, abi_platform)
+        lib_suffix = "t%s" % debug_suffix
```
jjhelmus: Would f-strings be better here and in similar changes?

jjhelmus:
In general this looks great but I am worried about the increase to CI time and the additional release artifacts associated with the build matrix expansion. I could see limiting the debug builds to just x86-64 or perhaps newer Python releases but I don't have a strong opinion here.

cpython-windows/build.py
vadikmironov: I converted the lines I touched in this PR, but there are ~150 more lines to touch. Do you want the whole file modernized? If yes, shall I do it here or perhaps under a separate PR?

vadikmironov:
Right, so the main CI impact will be on push-to-main job and not the PR build. If my calcs are right, the pipeline executes ~450 Linux jobs per push and additional 26 for the full suite would be ~5% increase. However, if I would to chose where to cut, I would not cut by architecture and rather limit the feature matrix. Happy to take a steer here - can drop debug or to do interleaved debug and debug+freethreaded or some other option whichever you prefer.

cpython-windows/build.py
jjhelmus: Just the touched lines is fine. Thanks.

jjhelmus:
Although the CI addition is small when considering the Linux jobs, the two are not equivalent. Windows runner are more expensive and currently these builds are the longest running especially the aarch64 Windows jobs which take an hour or more despite attempts to speed them up. Hence my concern about doubling the number of Windows builds.

jjhelmus:
I added labels that will trigger a more complete build matrix. I am closing + re-opening to trigger this.

jjhelmus (approved):

jjhelmus:
Because PGO is not needed the debug builds on Windows are much faster so the total CI time is not as bad as I was thinkings. I'm fine adding this to the build matrix, if the size causes problems down the line this can be reeevaluted by increasing the available runners or reducing the matrix.

vadikmironov:
thanks a lot for merging @jjhelmus. If you don't mind, I'll raise a %s uplift as a separate PR as it was bugging me to have mixed strings in single py script.

On the CI timings and PGO - with #1272 just merged (and you were totally right - it all works perfectly well), the native arm64 toolchain is now usable, supports PGO on windows aaaaand the build got a tiny bit faster (obv. PGO builds benefitting the most, but debug got faster too)!

Before and after this change:

| OS | Before, 3-run range | Spread | After (1 run) | Comments |
| --- | --- | --- | --- | --- |
| linux | 6150 to 6861 min | 711 | 5960 | unchanged |
| macos | 171 to 214 min | 43 | 170 | unchanged  |
| windows | 661 to 664 min | 4 | 1070 | +26 jobs, +410 min |

after #1272 with native Windows arm64 enabled, the 1070 dropped by 15% down to ~920 minutes. ok, not exactly a deal breaker and it's a tenth of what Windows debug matrix added, but many a mickle makes a muckle. Thanks again! 😄

PR #1263, opened 2026-09-12
Add debug support for MSVC builds - patch 1

vadikmironov:
This change is a part of set of changes addressing #26. This is exactly the same change as the first commit in #1262. It is raised separately for the ease of reviewing by maintainers while #1262  is to give an overview of the full set of patches (can be deleted once individual commits are merged).

What changes in `cpython-windows/build.py`:

- `debug` is added to the option set beside `noopt` and `pgo`, with the same ordering as Unix build already uses. Same as with Unix, this removes illegal combinations like `debug+pgo`. The set is renamed from `optimizations` to `options` also inline with the Unix build.
- `configuration="Debug"` is passed to msbuild and `artifact_config` accordingly. `PC/layout` gets `--debug` and that drives the `_d` artifact suffixes.
- executables, extension modules and dependency libraries pick up the `_d` suffix `$(PyDebugExt)` produces
- the tail-calling interpreter is not enabled for Debug (comments embedded in the patch)
- `docs/building.rst` updated with the debug option and the debug CRT requirement

Evidence: every debug configuration was built with this commit as the base, validated and tested. 26 jobs across 3.10 to 3.15 on x64, x86 and arm64, VS 2022 and VS 2026: https://github.com/vadikmironov/omniglot-bazel-starter/actions/runs/34095975591. The details, the archive layout and the other findings are in #1262.


jjhelmus:
Closing this as #1262 included this change.

PR #1264, opened 2026-09-15
[TESTING] try running arm64 build without emulation

jjhelmus:
Testing, no need to review.


jjhelmus:
Apparently modern releases of VS 2022 include a native arm64 compilers (rather than the x64 cross compiler for arm64 which is currently in use). This compiler can be selected with the (undocumented?) `vcvarsarm64.bat` script. Unfortunately this compiler does not support PGO so cannot be used.

PR #1265, opened 2026-09-15
Initialize Visual Studio env inside the build script

jjhelmus:
Move the `vcvarsall.bat` invocation inside the Python build script. 
Capture the resulting environment for use when building CPython and dependencies.

The target-triple is explicitly passes to the script or determined from the host Python interpreter. Previously this was derived from environment variables set prior to calling the script.


ci-targets.yaml
```diff
@@ -377,7 +377,6 @@ linux:
 windows:
   i686-pc-windows-msvc:
     arch: x86
-    vcvars: vcvars32.bat
```
charliermarsh, line 380: I think these are still referenced in `stdlib.yml` so the standard library tests would now fail!

charliermarsh (approved):
Approving but gotta fix the `stdlib.yml` thing before merging.

Also, do the [docs](https://github.com/astral-sh/python-build-standalone/blob/03f6d482736c52e0e768171fab064e0f31e4272f/docs/building.rst#L138-L163) need to be updated?

jjhelmus:
> Approving but gotta fix the stdlib.yml thing before merging.
Updated stdlib.yml in #1272 to avoid needing to rebase the stack.

I'll update docs in another PR.

PR #1266, opened 2026-09-15
Allow platform and compiler toolset versions to be specified in windows builds

jjhelmus:
Allow the platform and compiler toolset versions to be specified when performing builds on Windows. 

These components are updated with Visual Studio releases but older versions can be used with newer versions of Visual Studio. 

In particular when build CPython the v143 platform toolset is used for all builds except the 3.15 x86-64 build which used the v145 platform toolset. The corresponding compiler toolset versions are 14.30-14.45 and 14.50 - latest. By default the latest available toolset is activated by vcvarsall.bat but other toolsets can be selected via a command line argument.

This change decouples the Visual Studio version from the platform and compiler toolset version allowing the possibility of using newer Visual Studio versions to compile CPython and dependencies using older, compatible toolsets.


cpython-windows/build.py
```diff
@@ -1981,6 +2004,16 @@ def main() -> None:
...
     )
+    parser.add_argument(
+        "--platform-toolset",
+        choices=("v143", "v145"),
+        help="MSVC toolset (defaults to v145 for x64 CPython 3.15+, otherwise v143)",
+    )
+    parser.add_argument(
+        "--vc-tools-version",
+        help="MSVC version or prefix passed to vcvarsall "
+        "(defaults to 14.4 for v143 and 14.5 for v145)",
```
jjhelmus, line 2042: Note that 14.3 does **not** select the 14.4x toolsets. The logic seems to be simple directory prefix matching.

jjhelmus:
This is ready for review.

ci-targets.yaml
```diff
@@ -404,8 +406,12 @@ windows:
       - "3.14"
       - "3.15"
     vs_version: "2022"
-    vs_version_override_conditional:
+    platform_toolset: "v143"
+    vc_tools_version: "14.4"
+    windows_toolchain_override_conditional:
       vs_version: "2026"
+      platform_toolset: "v145"
+      vc_tools_version: "14.5"
```
charliermarsh, line 414: Why do we want to use this version for 3.15?

charliermarsh (approved):

ci-targets.yaml
jjhelmus: On x86-64 the v145 platform toolset that first ships with VS2026 is required to build a tail-calling interpreter which is configured in `cpython-windows/build.py`

PR #1268, opened 2026-09-15
[TESTING] test out VS2026 on Windows arm64

jjhelmus:
No need to review or merge this PR. It is only for testing and notes.

Test out migration to VS 2026 on windows-11-arm images.

A positive result using the `windows-11-vs2026-arm` was found after the change in 77a067d.


jjhelmus:
CPython builds certain build-time helpers, like `_freeze_module.exe` for the build machine which may differ from the target machine. This can be controlled via `/property:PreferredToolArchitecture=<architecture>`.

For the windows arm64 distributions, the build is done via a 32-bit x86 cross compiler running via emulation on arm64 hardware. As such these tools default to x86 rather which requires the corresponding Visual studio package install. The VS 2026 `windows-11-arm` image does not include a v143 x86 toolset, only an arm64 one (`Microsoft.VisualStudio.Component.VC.14.44.17.14.ARM64`). 

Patching the project file to build `_freeze_module.exe` for ARM64 remove the needs to a compatible x86 toolset without  effecting compiler and build tool selection.

PR #1271, opened 2026-09-16
Set MSBuild from Visual Studio environment PATH.

jjhelmus:
Use the MSBuild executable that is on PATH after activation of the Visual Studio environment.

Previously a hardcoded path was used that always selected a 32-bit x86 MSBuild executable.


charliermarsh (approved):
Makes sense!

PR #1272, opened 2026-09-16
Use a native arm64 compiler on Windows and migrate to VS 2026 image

jjhelmus:
Use a native arm64 tools to build the arm64 distributions on Windows. 
Previously 32-bit x86-cross-arm64 tools were used that ran under the Windows emulation layer on an arm64 host.
Migrate to a Windows arm images that uses VS 2026. The existing image is scheduled for migrate Sep 21-30.


jjhelmus:
The native arm64 compiler which is included in the VS 2022 install in the `windows-11-arm64` image does not support PGO. 

Based on the results in #1269 the tools included with VS 2026 in the `windows-11-vs2026-arm64` image do support PGO.

cpython-windows/build.py
charliermarsh, line 2027: Should this default to 2026 for ARM or something? Or reject PGO if ARM/2022? Not a big deal though.

charliermarsh (approved):
It looks like `libffi` is still built under emulation. We may want to change that in a follow-up?

Does building natively end up being faster, slower, or does it not make a big difference (with and without PGO)?

cpython-windows/build.py
jjhelmus: I think the requirement of VS 2026 for PGO using a native compiler is a result of the MSVC compiler version included in the `windows-11-arm64`. A latest v143 compatible MSVC available for VS 2022 can likely do a PGO although I cannot easily test this.

jjhelmus:
> Does building natively end up being faster, slower, or does it not make a big difference (with and without PGO)?

The build times are very noisy between runs so it is hard to get reliable numbers. 

In general the compilations steps of building CPython and OpenSSL are 10-60% faster with the native compilers. Other steps (prep, PGO training, compression) have no change. In total builds times drop by 10-20% but with significant variability between runs.

jjhelmus:
> It looks like libffi is still built under emulation. We may want to change that in a follow-up?
`prepare_libffi.bat` in the CPython source controls this. A patch could change this to use native compilers but I don't expect a significant improvements from this.

PR #1275, opened 2026-09-18
Use f-strings in the Windows build script

vadikmironov:
As a follow-up to #1262, this pr converts remainign 'str' cases into f-strings. Done via `ruff check --select UP031,UP032 --fix --unsafe-fixes cpython-windows/build.py` + %d and parenthesis cases manually.

`hack_props` `bytes` formats left untouched with `%`.


jjhelmus:
LGTM, thanks for the follow up.

PR #1276, opened 2026-09-19
Explicitly use Full LTO

jjhelmus:
CPython will uses ThinLTO in 3.12+ when is is supported (clang does). Explicitly select Full LTO for all builds for improved performance at the expense of longer build times and memory usage.


jjhelmus:
Python 3.14 built with Full LTO on Linux x86-64 show a ~1.3% improvement in `fastmark` benchmarks.
On macOS arm64 a 0.2% improvement was found.

Both of these results are rough, the machines were not isolated but the benchmarks were run with minimal other load.

indygreg (approved):
Man. CPython really needs a --with-release-optimizations or something if they are going to nerf the performance settings with the supposed _optimized_ build configuration.

